PR #1, opened 2018-07-17
fuse-overlayfs: try RENAME_EXCHANGE if the destination is a whiteout

giuseppe:
if we are trying to move over an existing whiteout, try a rename2(...,
RENAME_EXCHANGE) first, so in one atomic operation we get both the
file renamed and the whiteout created in the old source.


rhatdan:
LGTM
@rhvgoyal @nalind PTAL

giuseppe:
test are passing, I merge

PR #2, opened 2018-07-20
fuse-overlayfs: mountpoint is the last argument passed

giuseppe:
Consider the mountpoint to be the last argument passed, so that
fuse-overlayfs can be used as a mount helper.

This enables to run the github.com/amir73il/unionmount-testsuite
testsuite on fuse-overlayfs.


giuseppe:
Thanks!

PR #5, opened 2018-07-25
build: a couple of cleanups

giuseppe:
- fail if libfuse is not found.
- drop a bunch of files that are not used.


rhatdan:
LGTM

PR #14, opened 2018-08-01
fuse-overlayfs: honor trusted.overlay.origin

giuseppe:
While fuse-overlayfs doesn't set this xattr, ensure that we honor it when it exists so that we report the correct st_ino on lower layers that were created using the kernel implementation.


rhatdan:
@rhvgoyal PTAL

Issue #17, opened 2018-09-26
support for removing files


vbatts:
Built from 1c72a1af1a0c923907906260b1ab96d8d926949a

for lowerdir files, there is a whiteout needed when an unlink happens. Unlinking currently works for files created in the upperdir.

in one terminal:
```shell
[vbatts@poly] {master} ~$ mkdir /tmp/{upper,workdir,merged}
[vbatts@poly] {master} ~$ fuse-overlayfs -f -o lowerdir=$HOME,upperdir=/tmp/upper,workdir=/tmp/workdir /tmp/merged
UID=unchanged
GID=unchanged
UPPERDIR=/tmp/upper
WORKDIR=/tmp/workdir
LOWERDIR=/home/vbatts
MOUNTPOINT=/tmp/merged
```

in the other terminal:
```shell
[vbatts@poly] {master} /tmp/merged$ ls
bin/  Documents/  Downloads/  go1.10/  gopath/  memfetch.tgz  Projects/  src/  tmp/  tmp.984XhpMqlB/
[vbatts@poly] {master} /tmp/merged$ touch file 
[vbatts@poly] {master} /tmp/merged$ stat file
  File: file
  Size: 0               Blocks: 0          IO Block: 4096   regular empty file
Device: 46h/70d Inode: 2744322     Links: 1
Access: (0664/-rw-rw-r--)  Uid: ( 1000/  vbatts)   Gid: ( 1000/  vbatts)
Context: system_u:object_r:fusefs_t:s0
Access: 2018-09-26 11:07:09.558930372 -0400
Modify: 2018-09-26 11:07:09.558930372 -0400
Change: 2018-09-26 11:07:09.558930372 -0400
 Birth: -
[vbatts@poly] {master} /tmp/merged$ rm file
[vbatts@poly] {master} /tmp/merged$ rm memfetch.tgz
rm: cannot remove 'memfetch.tgz': Operation not supported
```

giuseppe:
I've opened a PR

vbatts:
this was on f28 `4.18.9-200.fc28.x86_64` kernel

PR #15, opened 2018-08-10
man: add man page

giuseppe:
Closes: https://github.com/containers/fuse-overlayfs/issues/7

Signed-off-by: Giuseppe Scrivano <gscrivan@redhat.com>

PR #19, opened 2018-10-05
fuse-overlayfs: try alternative whiteout on ENOTSUP

giuseppe:
if mknod fails with ENOTSUP then use the .wh.FILE variant.

Closes: https://github.com/containers/fuse-overlayfs/issues/17

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -14,6 +14,7 @@ addons:
       - xz-utils
       - python3.5
       - g++
+      - python3-setuptools
 before_install:
   - sudo mkdir -p /lower /upper /mnt
   - (cd /; sudo git clone https://github.com/amir73il/unionmount-testsuite.git)
--- a/main.c
+++ b/main.c
@@ -383,7 +383,7 @@ create_whiteout (struct ovl_data *lo, struct ovl_node *parent, const char *name,
       if (ret == 0)
         return 0;
 
-      if (errno != EPERM)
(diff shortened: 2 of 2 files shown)


rhatdan:
LGTM except tests are failing.

giuseppe:
the CI needed an additional package, tests are fine now.  @vbatts does this PR solve the issue you've seen?

vbatts:
something is not right with this now. I haven't even made it to fully testing #17, but just on the `cd /tmp/merged`.

Since I happen to have a large enough `$HOME` folder and a `$PS1` that looks for a .git folder, with this fix now `cd /tmp/merged` causes fuse-overlayfs to freeze when my $PS1 scans for git changes. I can't crtl+c, only `killall -9 fuse-overlayfs` and `umount /tmp/merged` to cleanup.

vbatts:
(this freeze does not happen in 79c70fd)

vbatts:
oh! but the freeze _does_ happen in 1c72a1a. So it isn't this PR, but a bisect between #13, #14, and #15

giuseppe:
I am still investigating, I've tried with some big git repositories I've under $HOME but I am still unable to reproduce the freeze.

The only patch that could introduce such issue seems to be b3898da6

Does it work for you if you revert that commit?

Also, what do you see if you use `--debug` or attach gdb/strace to it?  Where is it hanging?

giuseppe:
I am merging this PR as it also solves a problem with the CI

Issue #23, opened 2018-10-31
symlinks are broken


thiell:
Hi,
I was testing fuse-overlayfs when I noticed some issues when using symlinks. Sometimes symlinks stay in upper but are not visible from merged. It's also reproducible if you create a symlink that doesn't point to any existing file:
```
[sthiell@sh-101-60 /merged/lntest]$ ln -s non_existant_file link
[sthiell@sh-101-60 /merged/lntest]$ ls -l
total 0
[sthiell@sh-101-60 /merged/lntest]$ ls -lisa
total 8
144115205440289463 4 drwxr-xr-x 2 sthiell group 4096 Oct 31 14:12 .
144116351290596352 4 drwx--x--x 9 sthiell group 4096 Oct 31 14:12 ..
[sthiell@sh-101-60 /merged/lntest]$ ls -l /upper/lntest/
total 0
lrwxrwxrwx 1 root root 17 Oct 31 14:12 link -> non_existant_file
[sthiell@sh-101-60 /merged/lntest]$ cd ..
[sthiell@sh-101-60 /merged]$ rmdir lntest
rmdir: failed to remove ‘lntest’: Directory not empty
[sthiell@sh-101-60 /merged]$ rm -rf lntest
rm: cannot remove ‘lntest’: Directory not empty
```

giuseppe:
@thiell patch here: https://github.com/containers/fuse-overlayfs/pull/24

Could you verify if it solves the issue you are seeing?

thiell:
@giuseppe thanks, it works now! :)

PR #24, opened 2018-11-05
fuse-overlayfs: read correctly inode for dangling symlinks

giuseppe:
do not follow symlinks when stat'ing a path, so that we read the inode
of the symlink itself.

Closes: https://github.com/containers/fuse-overlayfs/issues/23

### DIFF
--- a/main.c
+++ b/main.c
@@ -1104,6 +1104,7 @@ do_lookup_file (struct ovl_data *lo, fuse_ino_t parent, const char *name)
     {
       int ret;
       char path[PATH_MAX];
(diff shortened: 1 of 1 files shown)


rhatdan:
LGTM

giuseppe:
let's merge this and iterate from here, I'll work on enabling tests for the `.wh.` variant

giuseppe:
new PR here to enable tests for the .wh. whiteout variant: https://github.com/containers/fuse-overlayfs/pull/25

PR #25, opened 2018-11-07
fuse-overlayfs: fix a couple of issues with the .wh. variant and enable tests

giuseppe:
before we were not testing the .wh. variant in Travis.

Enable tests and fix a couple of issues reported by the overlay testsuite

### DIFF
--- a/main.c
+++ b/main.c
@@ -83,6 +83,8 @@ struct _uintptr_to_must_hold_fuse_ino_t_dummy_struct
 };
 #endif
 
+static bool disable_ovl_whiteout;
+
 struct ovl_layer
 {
   struct ovl_layer *next;
@@ -374,7 +376,7 @@ create_whiteout (struct ovl_data *lo, struct ovl_node *parent, const char *name,
   int fd = -1;
   static bool can_mknod = true;
 
-  if (!skip_mknod && can_mknod)
+  if (!disable_ovl_whiteout && !skip_mknod && can_mknod)
     {
       int ret;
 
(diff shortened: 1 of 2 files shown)


main.c
```diff
@@ -83,6 +83,8 @@ struct _uintptr_to_must_hold_fuse_ino_t_dummy_struct
 };
 #endif
 
+static bool disable_ovl_whiteout;
```
rhatdan: Should this be initialized?

main.c
```diff
@@ -954,12 +957,18 @@ load_dir (struct ovl_data *lo, struct ovl_node *n, struct ovl_layer *layer, char
...
             }
 
+          sprintf (node_path, ".wh.%s", dent->d_name);
+          ret = TEMP_FAILURE_RETRY (fstatat (fd, node_path, &tmp_st, AT_SYMLINK_NOFOLLOW));
+          if (ret < 0 && errno != ENOENT)
+            {
+              closedir (dp);
+              return NULL;
+            }
           sprintf (node_path, "%s/%s", n->path, dent->d_name);
```
rhatdan: This looks strange?  You are overriding the previous settings of node_path?

main.c
giuseppe: non initialized global variables end in the `.bss` section.  It will be initialized to zero when the program starts

main.c
giuseppe: I am just re-using the same buffer so that I don't need more space on the stack.

rhatdan: OK, just checking.

rhatdan:
LGTM

PR #29, opened 2018-12-11
fuse-overlayfs: set umask to 0

giuseppe:
be sure we are able to set all the bits on created files/directories.


rhatdan:
LGTM

PR #30, opened 2018-12-14
copyup: fix error if lower file cannot be opened

giuseppe:
propagate the error code to the caller if the lower file cannot be
opened.


rhatdan:
LGTM

PR #32, opened 2018-12-20
fuse-overlayfs: set suid when running in a user namespace

giuseppe:
Addresses: https://github.com/containers/buildah/issues/1240#issuecomment-448949403


rhatdan:
LGTM

Issue #35, opened 2018-12-31
fusermount3: option allow_other only allowed if 'user_allow_other' is set in /etc/fuse.conf


siretart:
I'm trying to package this application for Debian (cf. http://bugs.debian.org/917889). While trying it out, I'm running into this issue:

```
./fuse-overlayfs -o lowerdir=/ -o upperdir=/tmp/upper/ -o workdir=/tmp/overlay-workdir/ /tmp/overlay
UID=unchanged
GID=unchanged
UPPERDIR=/tmp/upper
WORKDIR=/tmp/overlay-workdir/
LOWERDIR=/
MOUNTPOINT=/tmp/overlay
fusermount3: option allow_other only allowed if 'user_allow_other' is set in /etc/fuse.conf
./fuse-overlayfs: cannot mount: Operation not permitted
```

After editing `/etc/fuse.conf` to enable the option `user_allow_other`, the program appears to work. Unfortunately, this is not the default setting in Debian. My questions here are:

1. Is there any way to modify the program to work without editing `/etc/fuse.conf`?
2. What are the security ramifications for enabling this option?
3. Does RHEL8 enable this option by default? - If no, is the superuser expected to do that modification himself?

Thank you

giuseppe:
the `user_allow_other` is required to have multiple users inside of the container that can use the file system.  Otherwise fuse-overlayfs will be usable only by the user that is mapped to the unprivileged user on the host (usually root in the container).

siretart:
On Fri, Jan 4, 2019, 03:02 Giuseppe Scrivano <notifications@github.com
wrote:

> the user_allow_other is required to have multiple users inside of the
> container that can use the file system. Otherwise fuse-overlayfs will be
> usable only by the user that is mapped to the unprivileged user on the host
> (usually root in the container).
>

In my tests, that option is required for the program to work at all. Is
that expected or a bug?

giuseppe:
it is expected

rhatdan:
Should we document this in podman docs?

rhatdan:
@giuseppe Can you open a PR against Podman and here to document the issue.

siretart:
May I ask to clarify why this additional step is necessary? What are the security tradeoffs / implications by adding this option to `/etc/fuse.conf`?

I imagine the answer to this question should be included in the documentaiton.

giuseppe:
[...]

PR #39, opened 2019-01-09
fuse-overlayfs: use allow_other only when running with uid=0

giuseppe:
Closes: https://github.com/containers/fuse-overlayfs/issues/35


rhatdan:
@siretart PTAL

rhatdan:
LGTM

PR #42, opened 2019-01-30
setattr: use fd from fi->fh for ftruncate()

thiell:
When a file is created with restricted permissions (like 000), ftruncate() should
still work as long as the fd is open. Use fi->fh to fix that and avoid the call to
openat() which failed in that case.

### DIFF
--- a/main.c
+++ b/main.c
@@ -2593,22 +2593,33 @@ ovl_setattr (fuse_req_t req, fuse_ino_t ino, struct stat *attr, int to_set, stru
       fuse_reply_err (req, errno);
       return;
     }
-  if ((to_set & FUSE_SET_ATTR_SIZE))
+  if (to_set & FUSE_SET_ATTR_SIZE)
     {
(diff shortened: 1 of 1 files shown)


main.c
```diff
@@ -2593,22 +2593,16 @@ ovl_setattr (fuse_req_t req, fuse_ino_t ino, struct stat *attr, int to_set, stru
...
-      if (fd < 0)
-        {
-          fuse_reply_err (req, errno);
-          return;
-        }
+      int fd = fi->fh;
 
       if (ftruncate (fd, attr->st_size) < 0)
         {
           close (fd);
```
giuseppe: I think we must drop this line as well, we don't want to cleanup `fi->fh` in case of failure

thiell: Thanks @giuseppe! I was about to fix the close() issue, but I also noticed that in some case, fi might be NULL (https://libfuse.github.io/doxygen/structfuse__lowlevel__ops.html#a8fce08ae0d6a4f289bffbf9a9defe926). So I'll be pushing a new patch that can handle both cases. Let me know what you think.

giuseppe:
thanks a lot for the additional fix!  It looks good to me.

PR #45, opened 2019-02-15
fuse-overlayfs: use autocleanup functions

giuseppe:
first pass at using `__attribute__((cleanup))`.  It helps to simplify a lot of code, and hopefully avoid some bugs.

We should ideally add custom cleanup functions for the structs we are defining.

### DIFF
--- a/main.c
+++ b/main.c
@@ -2331,13 +2331,6 @@ ovl_do_open (fuse_req_t req, fuse_ino_t parent, const char *name, int flags, mod
       if (p == NULL)
         return -1;
 
-      if (delete_whiteout (lo, -1, p, name) < 0)
-        return -1;
-
-      sprintf (path, "%s/%s", p->path, name);
-      if (unlinkat (get_upper_layer (lo)->fd, path, 0) < 0 && errno != ENOENT)
-        return -1;
-
       sprintf (wd_tmp_file_name, "%lu", get_next_wd_counter ());
 
       fd = TEMP_FAILURE_RETRY (openat (lo->workdir_fd, wd_tmp_file_name, flags, mode & ~ctx->umask));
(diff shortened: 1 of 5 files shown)


main.c
```diff
@@ -2346,6 +2267,13 @@ ovl_do_open (fuse_req_t req, fuse_ino_t parent, const char *name, int flags, mod
           return -1;
         }
 
+      sprintf (path, "%s/%s", p->path, name);
```
rhatdan: Should this be snprintf?

rhatdan:
One question but the rest LGTM

main.c
giuseppe: good idea!  Now that we have autocleanup, we should use only dynamic allocated buffers.

I've added another patch to change all static buffers to dynamic allocated ones.

Please do not merge yet, I want to do some more testing as it is a lot of code changed, I will mark the PR as WIP.  If you are fine with the changes, I can merge after the stress tests

rhatdan: SGTM

giuseppe:
@rhatdan it is ready now

main.c
```diff
@@ -677,123 +718,157 @@ node_compare (const void *n1, const void *n2)
...
+
+      s = fgetxattr (sfd, name, buffer, current_size);
+      if (s < 0)
+        break;
+      if (s < current_size)
+        break;
+
+      current_size *= 2;
+      tmp = realloc (buffer, current_size + 1);
+      if (tmp == NULL)
```
rhatdan: Isn't this a potential leak of buffer here, on realloc failure.

main.c
```diff
@@ -1854,10 +1915,25 @@ copyup (struct ovl_data *lo, struct ovl_node *node)
...
+
+          ret = readlinkat (node_dirfd (node), node->path, p, current_size - 1);
+          if (ret < 0)
+            goto exit;
+          if (ret < current_size - 1)
+            break;
+
+          current_size = current_size * 2;
+          new = realloc (p, current_size);
+          if (new == NULL)
```
rhatdan: Need to free p on failure?

rhatdan:
A couple of questions on realloc, but nothing that I noticed wrong. Nice elimination of a lot of code.

main.c
giuseppe: `buffer` is marked with `cleanup_free`, so it is automatically freed when we leave the scope.

In facts, to avoid it is freed on success, we need to set it to NULL

main.c
giuseppe: same here, `p` is marked with `cleanup_free`, so it is freed anytime it is out of scope (and != NULL).

rhatdan:
LGTM

Issue #50, opened 2019-02-26
musl build failure


CameronNemo:
FTBFS with musl libc

```
cc -DHAVE_CONFIG_H -I.    -I . -I /builddir/fuse-overlayfs-0.3/lib -I/usr/include/fuse3 -fstack-clash-protection -D_FORTIFY_SOURCE=2 -mtune=generic -O2 -pipe    -MT fuse_overlayfs-main.o -MD -MP -MF .deps/fuse_overlayfs-main.Tpo -c -o fuse_overlayfs-main.o `test -f 'main.c' || echo './'`main.c
main.c:39:10: fatal error: error.h: No such file or directory
 #include <error.h>
          ^~~~~~~~~
compilation terminated.
make[2]: *** [Makefile:571: fuse_overlayfs-main.o] Error 1
```

https://travis-ci.org/void-linux/void-packages/jobs/498513645

CameronNemo:
My first thought is just adding a stub error() implementation. There are 20 uses of the function throughout the overlayfs module. Would you like me to take any special considerations into case when forming a patch?

giuseppe:
could you copy the implementation from gnulib?
It is already implemented there for systems where `error` is missing

giuseppe:
alternatively, I am fine if we take rid of `error` and replace with fprintf (stderr, ...) + exit ();

giuseppe:
@CameronNemo I've opened a PR here: https://github.com/containers/fuse-overlayfs/pull/52

mgoltzsche:
This is an issue again at least since `<error.h>` is used in [plugin_manager.c](https://github.com/containers/fuse-overlayfs/blob/b28a23a39d8b3b7346bf30c78ca758b9e30b0e2d/plugin-manager.c#L22) which is not supported by musl. Instead it should use `<err.h>` like [this](https://github.com/linux-can/can-utils/pull/163/files).

CameronNemo:
#147

PR #52, opened 2019-03-07
Fix musl build

giuseppe:
a few patches to address portability to musl libc

Closes: https://github.com/containers/fuse-overlayfs/issues/50


rhatdan:
LGTM

PR #53, opened 2019-03-08
fuse-overlayfs: fix lookup with multiple layers

giuseppe:
when looking up multiple layers and we have already found a file, do
not check if whiteouts are present in lower layers.  The lookup in the
lower layers is needed only to correctly propagate the inode number.


rhatdan:
LGTM

PR #55, opened 2019-03-27
fuse-overlayfs: fix first unlink done

giuseppe:
fix a regression introduced by f64f65287817fecd0 that prevents a
whiteout file to be created on the first unlink.


giuseppe:
@debarshiray this fixes the issue you've seen with unlink'ing `resolv.conf`

rhatdan:
LGTM

debarshiray:
Oh, nice!

I had eventually started bisecting it because I saw that https://github.com/debarshiray/toolbox/pull/34 or [toolbox commit 66e982af729a4b73](https://github.com/debarshiray/toolbox/commit/66e982af729a4b735d12f7c6e120c582efaea1f9) was also not working as intended, and had just narrowed it down to `fuse-overlayfs`. :)

Yes, it does fix the problems I was having.

debarshiray:
For what it's worth, fuse-overlayfs-0.3-4.dev.gitea72572 was the first Fedora build with the regression.

PR #56, opened 2019-03-28
docs: document the uid/gid mapping

giuseppe:


rhatdan:
LGTM

fuse-overlayfs.1.md
```diff
@@ -32,9 +32,58 @@ done to the file system will be written.
...
+65534 is the overflow id used when the UID/GID is not known inside the
+user namespace.  This happens because both users 0:0 and 1:1 are not
+mapped.
+
+The fuse-overlayfs dynamic mapping is an alternative and cheaper way
+to chown'ing the files on the host to accomodate the user namespace
+settings.
+
+In the above example, if we mount the fuse-overlayfs file system using:
+`-ouidmapping=0:1000:1:1:110000:65536,gidmapping=0:1000:1:1:110000:65536`,
```
DrDaveD: Isn't there supposed to be a space between "o" and "uidmapping?

fuse-overlayfs.1.md
```diff
@@ -32,9 +32,58 @@ done to the file system will be written.
...
+
+$ stat -c %u:%g merged/a merged/b
+65534:65534
+65534:65534
+
+65534 is the overflow id used when the UID/GID is not known inside the
+user namespace.  This happens because both users 0:0 and 1:1 are not
+mapped.
+
+The fuse-overlayfs dynamic mapping is an alternative and cheaper way
```
DrDaveD: I think this would be better if moved toward the beginning of the description, to help explain the purpose up front.

fuse-overlayfs.1.md
```diff
@@ -32,9 +32,58 @@ done to the file system will be written.
...
+In the above example, if we mount the fuse-overlayfs file system using:
+`-ouidmapping=0:1000:1:1:110000:65536,gidmapping=0:1000:1:1:110000:65536`,
+which is the namespace configuration specified on a single line, we'd
+see from the same user namespace:
+
+$ stat -c %u:%g merged/a merged/b
+0:0
+1:1
+
+That are the same IDs visible from outside the user namespace.
```
DrDaveD: Change "That" to "Those".

fuse-overlayfs.1.md
```diff
@@ -32,9 +32,58 @@ done to the file system will be written.
...
+without requiring to chown the files.
+
+For example, given on the host two files like:
+
+$ stat -c %u:%g lower/a lower/b
+0:0
+1:1
+
+When we run in a user namespace with the following configuration:
+$ cat /proc/self/uid_map
```
DrDaveD: Is it possible for fuse-overlayfs to read this itself instead of asking the user to pass it in in options?  Are the options just for overriding a default?

fuse-overlayfs.1.md
giuseppe: no it doesn't matter, `-o` is read as a single option

fuse-overlayfs.1.md
giuseppe: these are not really supposed to be used manually, the only user I am aware of right now is `containers/storage` so used internally by tools like Podman/Buildah/CRI-O.

Also, fuse-overlayfs itself is running outside of the namespace, so it is not enough to read `/proc/self/uid_map`

giuseppe:
thanks for the review, pushed a new version

fuse-overlayfs.1.md
DrDaveD: Ok, very interesting. I am thinking about how to use it as part of a [proposed Singularity feature for rootless containers](https://github.com/sylabs/singularity/pull/2718) along with kernel 4.18 or later.  I thought fuse-overlayfs would be used inside the namespace but I guess I don't yet fully understand it.

giuseppe:
@DrDaveD if you run it inside of the usernamespace, then it will already sees the "fake" IDs.

What we do in Podman for rootless containers is to create a usernamespace where all the mappings assigned to the user are present, i.e. the current user gets ID 0 then the additional subuids starts at 1.

Inside of that namespace, we can run different containers, each of them with their own sub mappings.

The fuse-overlayfs processes for the containers run into the first namespace, so that they have access to all the IDs.

If I can help with it, we could setup a meeting and it is probably easier to explain that here

Issue #57, opened 2019-03-28
Be less noisy


ihucos:
Hello,

maybe it makes sense to remove this output to stderr: https://github.com/containers/fuse-overlayfs/blob/a6958ce78fdabd75b7db4e9a6e03177b01734f8d/main.c#L3883

It is kind of noisy since its printed on every invocation and echoes parameters. Maybe only show it if some debugging flag is specified?

Cheers

ihucos

giuseppe:
agreed.  I've opened a PR to address it: https://github.com/containers/fuse-overlayfs/pull/58

PR #58, opened 2019-03-29
fuse-overlayfs: print config settings only with --debug

giuseppe:
Closes: https://github.com/containers/fuse-overlayfs/issues/57


rhatdan:
LGTM

PR #60, opened 2019-04-01
main: copyup uses reflinks if possible

giuseppe:
Use reflinks if they are supported by the underlying file system.

On my system, using XFS and reflinks, chowning recursively a directory
passed from:

User time (seconds): 0.02
System time (seconds): 0.34
Percent of CPU this job got: 8%
Elapsed (wall clock) time (h:mm:ss or m:ss): 0:04.52

to:

User time (seconds): 0.02
System time (seconds): 0.22
Percent of CPU this job got: 17%
Elapsed (wall clock) time (h:mm:ss or m:ss): 0:01.45

the performance improvement can be higher if the directory is made of
few but bigger files.

### DIFF
--- a/main.c
+++ b/main.c
@@ -36,6 +36,7 @@
 #include <assert.h>
 #include <errno.h>
 #include <err.h>
+#include <sys/ioctl.h>
 
 #ifdef HAVE_ERROR_H
 # include <error.h>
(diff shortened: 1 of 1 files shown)


rhatdan:
LGTM

giuseppe:
tests are passing now, merging..

Issue #59, opened 2019-03-31
Subtle errors when running from inside a chroot


ihucos:
When trying to run fuse-overlayfs from inside a chroot, there are hard to pin down, subtle errors.
This is some output of installing firefox inside arch linux with pacman.
```
(10/15) Creating temporary files...
Setting default ACL "u::rwx,g::r-x,g:wheel:r-x,g:adm:r-x,m::r-x,o::r-x" on /var/log/journal failed: Invalid argument
Setting access ACL "u::rwx,g::r-x,g:wheel:r-x,g:adm:r-x,m::r-x,o::r-x" on /var/log/journal failed: Invalid argument
Cannot set file attribute for '/var/log/journal', value=0x00800000, mask=0x00800000, ignoring: Function not implemented
Cannot set file attribute for '/var/log/journal/remote', value=0x00800000, mask=0x00800000, ignoring: Function not implemented
error: command failed to execute correctly
(11/15) Reloading device manager configuration...
  Skipped: Device manager is not running.
(12/15) Arming ConditionNeedsUpdate...
(13/15) Updating the info directory file...
(14/15) Updating the desktop file MIME type cache...
(15/15) Updating the MIME type database...
```
When using apt, there are errors about failed GPG checks, that I can not reproduce, it's kind of difficult to pin down.

This is interesting for nested containers. With `unionfs-fuse` I also get apparently the same kind of error: https://github.com/rpodgorny/unionfs-fuse/issues/84 . But much more often.

Let me know if I should try to find an reproducible example or give me some direction to investigate.

giuseppe:
are you running fuse-overlayfs as root?  How did you configure the chroot?

ihucos:
Hey giuseppe,

apparently I am mistaken, this has nothing to do with running it inside a chroot, but seems to be due unsupported operations:

This are the `strace`ed calls that caused the errors:
```
setxattr("/proc/self/fd/5", "system.posix_acl_access", "\2\0\0\0\1\0\7\0\377\377\377\377\4\0\5\0\377\377\377\377\10\0\5\0\346\3\0\0\10\0\5", 52, 0) = -1 EINVAL (Invalid argument)
```

```
openat(5, "log", O_RDONLY|O_NOFOLLOW|O_CLOEXEC|O_PATH) = 4
ioctl(4, FS_IOC_GETFLAGS, 0x7ffdfd0c0668) = -1 ENOSYS (Function not implemented)
```

```
[...]

PR #61, opened 2019-04-03
fuse-overlayfs: support some ioctls

giuseppe:
Closes: https://github.com/containers/fuse-overlayfs/issues/59

### DIFF
--- a/main.c
+++ b/main.c
@@ -3749,6 +3749,75 @@ ovl_fsync (fuse_req_t req, fuse_ino_t ino, int datasync, struct fuse_file_info *
   fuse_reply_err (req, ret == 0 ? 0 : errno);
 }
 
+static void
+ovl_ioctl (fuse_req_t req, fuse_ino_t ino, int cmd, void *arg,
+           struct fuse_file_info *fi, unsigned flags,
+           const void *in_buf, size_t in_bufsz, size_t out_bufsz)
+{
+  struct ovl_data *lo = ovl_data (req);
+  cleanup_close int cleaned_fd = -1;
+  struct ovl_node *node;
+  int fd = -1;
+  unsigned long r;
+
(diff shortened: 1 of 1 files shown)


rhatdan:
Does fuse call into this function?

rhatdan:
LGTM

giuseppe:
yes, it calls the `.ioctl` callback, that is registered to `ovl_ioctl`.

Let's wait for @ihucos comments before merging this

ihucos:
> Let's wait for @ihucos comments before merging this

Thanks. One thing that I don't understand in the whole picture, is that apparently calling `fuse-overlayfs` by root seems to work (see output in issue), in the sense that the program (`pacman`) using the `ioctl` does not complain in that case. How is that possible, if currently `ioctl` is not implemented in `fuse-overlayfs`.

About comments on the pull request I can only offer bikeshedding: One thing that at least in theory could surprise applications are maybe `errno` codes. As a developer I would only expect the `errno` codes listed in the man page of `ioctl`. I this pull request it could for example return errnos returned by `openat`, like `ENOMEM`, which would be passed as return value of ioctl, which at least in it's man page, does not have `ENOMEM` listed as possible error. I am kind of new to low level system programming and can't say how strict these usually are.

Very generally as a `fuse-overlayfs` user I would favor strictness over compatibility and prefer an application running on top of `fuse-overlayfs` to fail always than only sometimes.

giuseppe:
when it is running as root, if I try `chattr +i`, I get `chattr: Function not implemented while reading flags on merged/i`.  With the patch I get the correct behaviour.

The errno list in the man page is never exhaustive, there are many cases where syscalls can fail with errors not listed there.

ihucos:
LGTM :-)

PR #62, opened 2019-04-03
copyup: use sendfile(2) if available

giuseppe:
attempt to use sendfile(2) before falling back to a read/write loop.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -12,7 +12,7 @@ AC_PROG_CC
 gl_EARLY
 gl_INIT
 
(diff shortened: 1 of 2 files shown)


rhatdan:
LGTM
Which platform does not have sendfile?

giuseppe:
kernels older than 2.6.33 requires the destination to be a socket.  Don't think anyone is going to use fuse-overlayfs there, I've added mostly for safety if the code is copied somewhere else.

PR #63, opened 2019-04-04
ioctl: return immediately with not implemented ioctls

giuseppe:
avoid to send an error response twice.


rhatdan:
LGTM

PR #65, opened 2019-04-11
fuse-overlayfs: change ovl_statfs to use the upper layer fd instead o…

3XX0:
Use fstatvfs instead of statvfs, this is necessary if one `pivot_root` since the upperdir path won't be accessible.


rhatdan:
LGTM, but I will let @giuseppe Judge.

giuseppe:
thanks for the patch!  LGTM

PR #66, opened 2019-04-11
fuse-overlayfs: skip opaque whiteouts on unsupported filesystems (e.g…

3XX0:
Do not fail if xattr are not supported by the filesystem. This can happen for example if one uses a tmpfs as the upperdir.


giuseppe:
LGTM

PR #68, opened 2019-04-26
fuse-overlayfs: add support for .wh..wh..opq variant

giuseppe:
Closes: https://github.com/containers/libpod/issues/3021

### DIFF
--- a/main.c
+++ b/main.c
@@ -121,6 +121,7 @@ open_by_handle_at (int mount_fd, struct file_handle *handle, int flags)
 #define PRIVILEGED_XATTR_PREFIX "trusted.overlay."
 #define PRIVILEGED_OPAQUE_XATTR "trusted.overlay.opaque"
 #define PRIVILEGED_ORIGIN_XATTR "trusted.overlay.origin"
+#define OPAQUE_WHITEOUT ".wh..wh..opq"
 
 #if !defined FICLONE && defined __linux__
 # define FICLONE _IOW (0x94, 9, int)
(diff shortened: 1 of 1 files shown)


main.c
```diff
@@ -1121,14 +1135,14 @@ load_dir (struct ovl_data *lo, struct ovl_node *n, struct ovl_layer *layer, char
...
 
-          ret = TEMP_FAILURE_RETRY (fstatat (fd, whiteout_path, &tmp_st, AT_SYMLINK_NOFOLLOW));
-          if (ret < 0 && errno != ENOENT)
-            return NULL;
-
           ret = asprintf (&node_path, "%s/%s", n->path, dent->d_name);
           if (ret < 0)
             return NULL;
 
+          ret = TEMP_FAILURE_RETRY (fstatat (fd, whiteout_path, &tmp_st, AT_SYMLINK_NOFOLLOW));
```
rhatdan: Any reason to move this?

main.c
```diff
@@ -417,15 +417,20 @@ has_prefix (const char *str, const char *pref)
...
     {
       if (errno == ENOTSUP)
-        return 0;
+        goto create_opq_whiteout;
       if (errno != EPERM || fsetxattr (fd, OPAQUE_XATTR, "y", 1, 0) < 0 && errno != ENOTSUP)
           return -1;
     }
-
-  return 0;
+ create_opq_whiteout:
```
rhatdan: It looks like you are creating the opq_whiteout even if the fsetxattr above is successful?  Is this intentional

rhatdan:
One question, if the answer was that is intentional then LGTM

main.c
giuseppe: yes, I am not sure we still want the `OPAQUE_XATTR` which is fuse-overlayfs specific.  I've not removed it yet as it might break existing images.

PR #70, opened 2019-05-02
copyup: minor fixes for sendfile()

thiell:


main.c
```diff
@@ -2093,15 +2093,16 @@ copyup (struct ovl_data *lo, struct ovl_node *node)
...
+          ssize_t n = TEMP_FAILURE_RETRY (sendfile (dfd, sfd, NULL, tocopy > SIZE_MAX ? SIZE_MAX : (size_t) tocopy));
           if (n < 0)
             {
               /* On failure, fallback to the read/write loop.  */
               ret = copy_fd_to_fd (sfd, dfd, buf, buf_size);
               if (ret < 0)
                 goto exit;
             }
-          copied += n;
+          else
```
giuseppe: not introduced by this PR, but since we are here, could we break the loop once we exit from copy_fd_to_fd?  I think I missed a `break;` after the `goto exit;`

giuseppe:
thanks a lot!  LGTM

Issue #69, opened 2019-05-01
Segfault on rename (?)


aengelke:
Apparently, fuse-overlayfs can segfault on rename under some (to be identified) condition.

Reproducer on current Silverblue (F30, commit `06ef84d208a1c29ce18a00b51d4ef5cd101365037f580cde0bbed423b2c051e3`):

1. Create and enter some container (in my case, `toolbox create` and `toolbox enter`)
2. Run `sudo rm /usr/share/glib-2.0/schemas/gschemas.compiled`
2. Run `sudo glib-compile-schemas /usr/share/glib-2.0/schemas/`
3. The final rename fails and `Bus error` is written.

```
$ sudo glib-compile-schemas /usr/share/glib-2.0/schemas/
Warning: Schema “org.gnome.system.locale” has path “/system/locale/”.  Paths starting with “/apps/”, “/desktop/” or “/system/” are deprecated.
Warning: Schema “org.gnome.system.proxy” has path “/system/proxy/”.  Paths starting with “/apps/”, “/desktop/” or “/system/” are deprecated.
Warning: Schema “org.gnome.system.proxy.http” has path “/system/proxy/http/”.  Paths starting with “/apps/”, “/desktop/” or “/system/” are deprecated.
Warning: Schema “org.gnome.system.proxy.https” has path “/system/proxy/https/”.  Paths starting with “/apps/”, “/desktop/” or “/system/” are deprecated.
Warning: Schema “org.gnome.system.proxy.ftp” has path “/system/proxy/ftp/”.  Paths starting with “/apps/”, “/desktop/” or “/system/” are deprecated.
Warning: Schema “org.gnome.system.proxy.socks” has path “/system/proxy/socks/”.  Paths starting with “/apps/”, “/desktop/” or “/system/” are deprecated.
Failed to rename file “/usr/share/glib-2.0/schemas/gschemas.compiled.49TX0Z” to “/usr/share/glib-2.0/schemas/gschemas.compiled”: g_rename() failed: Software caused connection abort
Bus error
```

From the logs:

```
May 01 12:38:48 nuc audit[7529]: ANOM_ABEND auid=1000 uid=1000 gid=1000 ses=2 subj=unconfined_u:system_r:container_runtime_t:s0-s0:c0.c1023 pid=7529 comm="fuse-overlayfs" exe="/usr/bin/fuse-overlayfs" sig=11 res=1
[...]

PR #71, opened 2019-05-03
fuse-overlayfs: do not copyup a whiteout on rename

giuseppe:
Closes: https://github.com/containers/fuse-overlayfs/issues/69


rhatdan:
LGTM

PR #72, opened 2019-05-11
Update static build instructions

sVattic:
There were three problems with the static build instructions:

1. `buildah` can't find an image without a search repository configured
1. I was getting a build error about missing compilers until I added clang
1. The readme had the `-t` and `-f` flags mixed up

With these changes, I was able to build with buildah `1.9.0-dev` :tada:

### DIFF
--- a/Dockerfile.static
+++ b/Dockerfile.static
@@ -1,7 +1,7 @@
-FROM fedora:latest
+FROM docker.io/fedora:latest
 WORKDIR /build
 RUN dnf update -y && \
-    dnf install -y git make automake autoconf gcc glibc-static meson ninja-build
+    dnf install -y git make automake autoconf gcc glibc-static meson ninja-build clang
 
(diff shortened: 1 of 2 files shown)


Dockerfile.static
```diff
@@ -1,7 +1,7 @@
-FROM fedora:latest
+FROM docker.io/fedora:latest
```
rhatdan: I would prefer
registry.fedoraproject.org/fedora:latest
docker.io is just a redirect to fedora in this case, I believe
Does this work for you?

yuvipanda:
Hah, I didn't notice this when I filed #74, #72 and #75. I'll kill them all.

giuseppe:
LGTM with the change @rhatdan requested.  Could you please amend it?

sVattic:
OK, I've updated the Dockerfile with @rhatdan's changes and confirmed that it still works.

rhatdan:
Thanks @sVattic 
LGTM

Issue #73, opened 2019-05-12
Static build fails


yuvipanda:
<3 project.

Static build fails with:

```
 ---> Running in 5ce3b3418daf
Cloning into 'libfuse'...
The Meson build system
Version: 0.50.0
Source dir: /build/libfuse
Build dir: /build/libfuse/build
Build type: native build
Project name: libfuse3
Project version: 3.5.0
Appending LDFLAGS from environment: '-lpthread'
Appending LDFLAGS from environment: '-lpthread'
Native C compiler: cc (gcc 8.3.1 "cc (GCC) 8.3.1 20190223 (Red Hat 8.3.1-2)")

meson.build:1:0: ERROR: Unknown compiler(s): ['c++', 'g++', 'clang++', 'pgc++']
The follow exceptions were encountered:
Running "c++ --version" gave "[Errno 2] No such file or directory: 'c++': 'c++'"
Running "g++ --version" gave "[Errno 2] No such file or directory: 'g++': 'g++'"
Running "clang++ --version" gave "[Errno 2] No such file or directory: 'clang++': 'clang++'"
Running "pgc++ --version" gave "[Errno 2] No such file or directory: 'pgc++': 'pgc++'"

A full log can be found at /build/libfuse/build/meson-logs/meson-log.txt
The command '/bin/sh -c git clone https://github.com/libfuse/libfuse &&     cd libfuse &&     mkdir build &&     cd build &&     LDFLAGS="-lpthread" meson --prefix /usr -D default_library=static .. &&     ninja &&     ninja install' returned a non-zero code: 1
```

yuvipanda:
Fixed with #74

yuvipanda:
Done in #72

PR #75, opened 2019-05-12
Pin fedora image used for static build

yuvipanda:
Latest as of today. Makes Dockerfile more reproducible
in the long run, helps pin down build failures more easily.

When investigating #73, I wasn't sure if the issue was
that `fedora:latest` has changed since last commit to
Dockerfile.static, or something else. Pinning the image
makes makes it easier to reason about 'did a source
change or did a base image change break this?'


yuvipanda:
Done in #72

PR #76, opened 2019-05-15
main: skip ENOTDIR in a lookup

giuseppe:
if we are looking for a path in a lower component, skip ENOTDIR as a
component might be a whiteout.

This happens only when running as root, as the whiteout takes the same
name of the deleted file/directory.


rhatdan:
LGTM

PR #78, opened 2019-05-24
fuse-overlayfs: some cleanups

giuseppe:
these are just some cleanup patches that I've prepared while working on a different feature:

https://github.com/containers/fuse-overlayfs/compare/master...giuseppe:multiple-sources?expand=1


rhatdan:
LGTM

PR #79, opened 2019-05-27
[WIP] fuse-overlayfs: add plugin system

giuseppe:
support loading layers on demand.

Add a simple plugin mechanism that will help to expand fuse-overlayfs
functionalities, in particular it allows to load data from a layer on
demand.

A plugin is loaded into fuse-overlayfs using the option:

-o plugins=path/to/plugin.so:path/to/another/plugin.so

A layer can use a plugin with the syntax:

-o lowerdir=//plugin-name/DATA-FOR-THE-PLUGIN/path

Each time a file/directory is looked up, if a plugin is registered for
a layer, the plugin is first notified about the request.

After the callback is invoked, fuse-overlayfs still expects the data
to be accessible at the specified directory.

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -4,14 +4,14 @@ bin_PROGRAMS = fuse-overlayfs
 
 ACLOCAL_AMFLAGS = -Im4
 
-EXTRA_DIST = m4/gnulib-cache.m4 rpm/fuse-overlayfs.spec.template autogen.sh fuse-overlayfs.1.md utils.h NEWS
+EXTRA_DIST = m4/gnulib-cache.m4 rpm/fuse-overlayfs.spec.template autogen.sh fuse-overlayfs.1.md utils.h plugin.h NEWS
 
 CLEANFILES = fuse-overlayfs.1
 
 fuse_overlayfs_CFLAGS = -I . -I $(abs_srcdir)/lib $(FUSE_CFLAGS)
 fuse_overlayfs_LDFLAGS =
 fuse_overlayfs_LDADD = lib/libgnu.a $(FUSE_LIBS)
-fuse_overlayfs_SOURCES = main.c
+fuse_overlayfs_SOURCES = main.c plugin.c
 
 WD := $(shell pwd)
 
--- a/configure.ac
+++ b/configure.ac
@@ -54,6 +54,7 @@ LDFLAGS=$old_LDFLAGS
 
 AC_DEFINE_UNQUOTED([HAVE_FUSE_CACHE_READDIR], $cache_readdir, [Define if libfuse is available])
 
+AC_SEARCH_LIBS([dlopen], [dl], [], [AC_MSG_ERROR([unable to find dlopen()])])
 
 AC_FUNC_ERROR_AT_LINE
 AC_FUNC_MALLOC
--- a/main.c
+++ b/main.c
@@ -79,6 +79,8 @@
 
 #include <pthread.h>
 
+#include <plugin.h>
+
 #ifndef TEMP_FAILURE_RETRY
 #define TEMP_FAILURE_RETRY(expression) \
   (__extension__                                                              \
@@ -188,6 +190,9 @@ struct ovl_layer
   char *path;
   int fd;
   bool low;
+
+  struct ovl_plugin *plugin;
+  void *plugin_opaque;
 };
 
 struct ovl_mapping
@@ -239,6 +244,7 @@ struct ovl_data
   char *gid_str;
   struct ovl_mapping *uid_mappings;
   struct ovl_mapping *gid_mappings;
+  char *plugins;
   char *mountpoint;
   char *lowerdir;
   char *context;
@@ -254,6 +260,7 @@ struct ovl_data
   struct ovl_node *root;
   char *timeout_str;
   double timeout;
+
   int threaded;
   int fsync;
   int fast_ino_check;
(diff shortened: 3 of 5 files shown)


giuseppe:
to give some context, I am using it to play with it: https://github.com/giuseppe/crfs/tree/fuse-overlay-plugin.

I've prepared a stargz fedora image and pushed it to `quay.io/giuseppe/fedora`.

```
## using the fuse-overlay-plugin branch
$ (cd $GOPATH/src/github.com/giuseppe/crfs; make -C plugin/ && go build)
$ export PATH=$GOPATH/src/github.com/giuseppe/crfs:$PATH
$ mkdir lower upperdir work merged
$ fuse-overlayfs -oplugins=$GOPATH/src/github.com/giuseppe/crfs/plugin/crfs.so,lowerdir=//crfs/$(echo -n https://quay.io/v2/giuseppe/fedora/blobs/sha256:376969999a023ffbd0de97eddc3b87b2f868b3ae71a43cc7c4de39438383b608 | base64 -w0)/lower,upperdir=upper,workdir=work merged
$ time podman run --rm -ti --rootfs $(pwd)/merged /usr/bin/sh -c 'echo hello world'
hello world

real    0m3.684s
user    0m0.106s
sys     0m0.037s
```

rhatdan:
Very cool, could we make this a feature of Podman?  IE Push in Star.tgz format?

giuseppe:
yes, it is not difficult to generate the tarballs in the star.tgz format.

What is difficult is to add the concept of read-on-the-fly layers in our tools.  In any case, before that, we should investigate better if it has any significant improvement.  Using the `fedora` image, I don't see any clear advantage so far.

rhatdan:
Do we need something far larger, like one of the jboss container images

bradfitz:
@giuseppe, I haven't done a number of HTTP optimizations in the crfs code yet. Its access patterns are still pretty poor, no readahead, no caching of requests already done, etc. I was waiting to get it working end-to-end before optimizing those.

giuseppe:
> @giuseppe, I haven't done a number of HTTP optimizations in the crfs code yet. Its access patterns are still pretty poor, no readahead, no caching of requests already done, etc. I was waiting to get it working end-to-end before optimizing those.

@bradfitz yes I agree there is still much margin for improvements and I am optimist this can be helpful for different use cases :-)

The plugin mechanism I've added in this PR was the simplest implementation I could think of to allow me to play directly with CRFS.  It works fine, I've tested a few images and could not see any issue.  The ugliest part is that a `readdir(x)` requires to pull also every child of `x`.  We could delegate each FUSE request to the plugin/driver, but that requires a much more significant refactoring, and also introduces some other problems (inode numbers and passing FDs so that read/write operations can use `splice(2)`.

From a quick profiling, most of the CPU time is spent in the gzip decompression.  We are already using an alternative version in our libraries (`klauspost/pgzip`) that significantly improves the decompression time on the single stream, but in this case I've seen it performing worse than `compress/gzip`.  I had no time yet to try it out, but I'd like to see how `facebook/zstd` performs with container images.  Unfortunately this is a breaking change for existing clients, while stargz is not.

In the last version I am working on, I've added a readahead so that the file is fetched also while there are not active requests: https://github.com/giuseppe/crfs/blob/537cd10cae39662cdbf8dbcd688894bc53aa124c/crfs.go#L284-L301
(For our container tools it could also make sense in future to have a mode to store the image once the layers are fully downloaded).

The generic use case I was trying, `yum upgrade` on a Fedora image, is probably an extreme test case, as yum needs to load many tiny python files.  Golang binaries will surely perform better.

@rhatdan I can try a jboss-like image and see how it performs

giuseppe:
@bradfitz I've noticed a huge performance improvement if I use [zstd](https://github.com/facebook/zstd) instead of gzip.  I've opened two PRs so that our containers tools understand the zstd compression:

https://github.com/containers/image/pull/639
https://github.com/containers/storage/pull/363

AkihiroSuda:
@giuseppe Using zstd requires defining new OCI media type for distribution?

giuseppe:
> @giuseppe Using zstd requires defining new OCI media type for distribution?

from https://github.com/containers/image/pull/639#issuecomment-505473848 it seems it is enough to add the `+zstd` suffix to the mediaType.  It would be nice to have the support for zstd also in Docker.

From some quick tests, there are significant improvements also when not using something like `stargz` (`starzstd`? :) ): https://github.com/containers/image/pull/639#issuecomment-517669404

AkihiroSuda:
any plan toward merging this? ^^

giuseppe:
> any plan toward merging this? ^^

I had no time to look at this recently.  I've still some doubts about it as the current implementation for crfs using gzip is quite slow, that is why I have started looking into zstd (PR still blocked on review for containers/image) as a way to speed it up.

rhatdan:
@giuseppe I just merged your zstd PR in containers/image

giuseppe:
> @giuseppe I just merged your zstd PR in containers/image

thanks, I've opened a PR for `opencontainers/image-spec` to add the zstd compression: https://github.com/opencontainers/image-spec/pull/788

giuseppe:
there are other use cases I'd like to play with in fuse-overlayfs, so I've worked on an alternative version for plugins that offer more flexibility: https://github.com/containers/fuse-overlayfs/pull/119

giuseppe:
I am closing this PR in favour of the new implementation.

Unfortunately I'll need to adapt the CRFS PoC I had worked with as it is broken now with the new API

Issue #80, opened 2019-05-30
Multiple flock calls on the same filename deadlocks


basvdlei:
When flock calls are made from multiple processes for the same filename fuse-overlayfs deadlocks.

This happens reliably when running multiple `git clones` or `go get` from a rootless container in Fedora Silverblue:

```
● ostree://fedora-workstation:fedora/30/x86_64/silverblue
                   Version: 30.20190530.0 (2019-05-30T00:39:27Z)
                BaseCommit: 8980028d1ac8f8180b29233be4704d0ff3c8e01d6833d149dc460919a63c39a7
```

Backtrace of hanging `/usr/bin/fuse-overlayfs`:
```c
#0  0x00007f68d23a647b in flock () from target:/lib64/libc.so.6
#1  0x00000000004031be in ovl_flock ()
#2  0x00007f68d24b71ba in ?? () from target:/lib64/libfuse3.so.3
#3  0x00007f68d24b9908 in ?? () from target:/lib64/libfuse3.so.3
#4  0x00007f68d24b4c2b in fuse_session_loop () from target:/lib64/libfuse3.so.3
#5  0x0000000000402c56 in main ()
```

It's probably related to the following section in the flock(2) manpage:

> If a process uses open(2) (or similar) to obtain more than one file descriptor for the same file, these file descriptors are treated independently by flock(). *An attempt to lock the file using one of these file descriptors may be denied by a lock that the calling process has already placed via another file descriptor.*

Maybe related to #67 

## How to reproduce

Running the following program in a rootless container concurrently hangs that container.

**fuse-overlay-flock.c**
```c
#include <sys/file.h>
#include <stdio.h>
#include <unistd.h>

int main(int argc, char* argv[])
{
	int fd;
	fd = open(argv[1], O_CREAT|O_WRONLY, 0644);
	flock(fd, LOCK_EX);
	dprintf(fd, "foo\n");
	flock(fd, LOCK_UN);
	close(fd);
	printf("done\n");
	return 0;
}
```

**Dockerfile**:
```dockerfile
FROM registry.fedoraproject.org/fedora:30
RUN dnf install -y gcc
COPY fuse-overlay-flock.c /
RUN gcc -o fuse-overlay-flock fuse-overlay-flock.c
ENTRYPOINT [ "/usr/bin/bash" ]
CMD [ "-c", "for i in $(seq 1 10); do ( /fuse-overlay-flock testfile & ); done" ]
````

**Command**:
```bash
[...]

PR #82, opened 2019-06-03
main: drop code to handle flock

giuseppe:
the implementation of flock(2) was wrong and it caused a deadlock when
multiple processes tried to lock the same file.

Drop it and let FUSE handle locking for us.

Closes: https://github.com/containers/fuse-overlayfs/issues/80


rhatdan:
LGTM

PR #83, opened 2019-06-04
fuse-overlayfs: add new option "timeout"

giuseppe:
allow to override the default timeout value we specify for an entry
validity.


rhatdan:
LGTM

PR #88, opened 2019-07-15
fuse-overlayfs: different performance tweaks

giuseppe:
this PR has different performance improvements, most notably, there is some initial support for threading.  Most of the code is still protected by a lock, so it runs on a single threaded, but operations like read/write/setattr/fsync can be dispatched on a different thread.

It is now possible to disable `fsync` that seems to slow `apt` quite significantly.  In general it is probably safer to run containers without access to fsync.

The new options can be configured in Podman through `~/.config/containers/storage.conf`.  I've got the  best results with this combination:

`mountopt = "threaded=0,fsync=0"`

PR #89, opened 2019-07-23
fuse-overlayfs: different performance tweaks continuation of #88 

giuseppe:
there are some other performance fixes.

In particular, this PR addresses some issues reported with: https://github.com/containers/fuse-overlayfs/pull/88#issuecomment-513882925

Now operations like readdir/rmdir/unlink require significantly less syscalls.


rhatdan:
@ mbargull  FYI
LGTM

PR #97, opened 2019-07-31
main: fix type for flistxattr return

giuseppe:
the return type is ssize_t, not size_t.


rhatdan:
LGTM

Issue #99, opened 2019-08-06
v0.5 breaks postgres in rootless podman


neVERberleRfellerER:
Postgres does not longer run properly in rootless podman with fuse-overlayfs v0.5. It works with v0.4.1.
Now it fails with
```
waiting for server to start....2019-08-06 16:39:14.659 UTC [44] LOG:  could not set permissions of file "/var/run/postgresql/.s.PGSQL.5432": No such device or address
2019-08-06 16:39:14.659 UTC [44] WARNING:  could not create Unix-domain socket in directory "/var/run/postgresql"
2019-08-06 16:39:14.659 UTC [44] FATAL:  could not create any Unix-domain sockets
```

How to reproduce:
```
podman run --rm -it postgres
```

This is on
`Linux linux1 5.2.6-arch1-1-ARCH #1 SMP PREEMPT Sun Aug 4 14:58:49 UTC 2019 x86_64 GNU/Linux`
But I tested it with older kernels too (some 5.1.x)

**UPDATE**
Poor mans bisect with makepkg shows this:
```
1395753e1a17da68f35ab05b805878b9da0972b1 broken
9e20d96bbf2de0f32abb7ab8eb9bd02f11510c9e broken
62cbd9977c5be5968e0c1d404c26c5a346a486f2 broken
0b0c7a3a01dae4de65ba79016a111d794c1d1719 broken
468cb46773dc70349d511e1f5d4748d792e979b3 ok
241fecf92b9540266affb950bcf29866bcd0167c ok
d092552b76732bbf2f2f0bdbd4cf458f13ded330 ok
eb509974efe9e21a122a93a6cc16379ae66ef288 ok
d9c5215e33f3cee938b263b80d3c8f628234b9e5 ok
```
so https://github.com/containers/fuse-overlayfs/commit/0b0c7a3a01dae4de65ba79016a111d794c1d1719 it is .

giuseppe:
thanks a lot for the detailed report.

I've just opened a PR, could you double check if it solves the issue you've seen: https://github.com/containers/fuse-overlayfs/pull/101?

neVERberleRfellerER:
>  could you double check if it solves the issue you've seen: #101?

It does. Thank you.

giuseppe:
thanks for confirming it so quickly.  I had to push another version as the previous version was breaking some tests.

neVERberleRfellerER:
Latest version of #101 stil works.

rhatdan:
[...]

PR #101, opened 2019-08-07
main: simplify setattr

giuseppe:
always use the /proc/self/fd/FD path when a fd is not available.

commit 0b0c7a3a01dae4de65ba79016a111d794c1d1719 introduced the regression.

Closes: https://github.com/containers/fuse-overlayfs/issues/99


rhatdan:
LGTM

PR #103, opened 2019-08-11
main: use cached st_mode

giuseppe:
node already keeps st_mode in cache so there is no need of the
additional stat to find out the file type.


rhatdan:
LGTM

PR #105, opened 2019-08-15
main: if the path is a dir, work on the proc path

giuseppe:
when reading xattr, if the path cannot be open as it is a directory, operate on the /proc/fd/FD path.

Closes: https://github.com/containers/fuse-overlayfs/issues/104


rhatdan:
LGTM

Issue #108, opened 2019-08-21
dnf transaction fails inside Fedora container and corrupts container


yann-soubeyrand:
/kind bug

**Description**

Doing a dnf install inside a Fedora container or a Silverblue toolbox results in failed transaction and corrupted container.

**Steps to reproduce the issue:**

1. podman --log-level=debug run -ti --restart=no --rm fedora

2. Inside container: dnf install jq (jq is an example, it could be dnf upgrade)

**Describe the results you received:**

DNF transaction failed. If the command used is dnf upgrade the system is corrupted (library files from the packages in the transaction are deleted).

```
INFO[0000] running as rootless                          
DEBU[0000] Initializing boltdb state at /var/home/yann/.local/share/containers/storage/libpod/bolt_state.db 
DEBU[0000] Using graph driver overlay                   
DEBU[0000] Using graph root /var/home/yann/.local/share/containers/storage 
DEBU[0000] Using run root /tmp/1000                     
DEBU[0000] Using static dir /var/home/yann/.local/share/containers/storage/libpod 
DEBU[0000] Using tmp dir /run/user/1000/libpod/tmp      
DEBU[0000] Using volume path /var/home/yann/.local/share/containers/storage/volumes 
DEBU[0000] Set libpod namespace to ""                   
DEBU[0000] [graphdriver] trying provided driver "overlay" 
DEBU[0000] overlay: mount_program=/usr/bin/fuse-overlayfs 
DEBU[0000] backingFs=extfs, projectQuotaSupported=false, useNativeDiff=false, usingMetacopy=false 
DEBU[0000] Initializing event backend journald          
DEBU[0000] parsed reference into "[overlay@/var/home/yann/.local/share/containers/storage+/tmp/1000:overlay.mount_program=/usr/bin/fuse-overlayfs]docker.io/library/fedora:latest" 
DEBU[0000] parsed reference into "[overlay@/var/home/yann/.local/share/containers/storage+/tmp/1000:overlay.mount_program=/usr/bin/fuse-overlayfs]@ef49352c9c21ca58ce753fd0b5dda645492236cec4213ac31ee47c35f2e91b1c" 
DEBU[0000] exporting opaque data as blob "sha256:ef49352c9c21ca58ce753fd0b5dda645492236cec4213ac31ee47c35f2e91b1c" 
[...]

PR #109, opened 2019-08-22
main: introduce inodes cache

giuseppe:
introduce a hash map to refer from an inode to the file paths.
A recent change where we enable FUSE writeback by default uncovered an
underlying issue in fuse-overlayfs where changes to a file with
multiple links would not be visible from the other link.

For each inode, maintain a list of nodes that refer to it, so that we
can still access it when a link is removed and more importantly we can
use the inode value with FUSE.

Closes: https://github.com/containers/fuse-overlayfs/issues/108
Closes: https://bugzilla.redhat.com/show_bug.cgi?id=1744109

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -4,7 +4,7 @@ bin_PROGRAMS = fuse-overlayfs
 
 ACLOCAL_AMFLAGS = -Im4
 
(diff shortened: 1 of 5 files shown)


giuseppe:
tests are passing

giuseppe:
@rhatdan if you are fine with this PR, I'll cut a new version

rhatdan:
LGTM

Issue #112, opened 2019-08-27
NULL node->children passed to hash_lookup


nwf:
Using `fuse-overlayfs` to back a rootless container, I seem to be able to reliably trigger the following crash in `v0.6.1`.  The `make` process inside the container runs
```
450   openat(AT_FDCWD, "/opt/build/build_dir/target-arm_xscale_musl_eabi/linux-kirkwood/linux-4.14.138/.vermagic", O_RDONLY <unfinished ...>
```

and `fuse-overlayfs` dies with

```
Program received signal SIGSEGV, Segmentation fault.
0x000000010ad12e40 in safe_hasher (table=table@entry=0x0, key=key@entry=0x7fffe618ba20) at hash.c:251                                                       
251       size_t n = table->hasher (key, table->n_buckets);
(gdb) bt
#0  0x000000010ad12e40 in safe_hasher (table=table@entry=0x0, key=key@entry=0x7fffe618ba20) at hash.c:251                                                   
#1  0x000000010ad13688 in hash_lookup (table=0x0, entry=0x7fffe618ba20) at hash.c:263                                                                       
#2  0x000000010ad0b684 in do_lookup_file (lo=0x7fffe618bf50, parent=<optimized out>, name=0x7fffbe1c0038 "toolchain-arm_xscale_gcc-7.4.0_musl_eabi")        
    at main.c:1739
#3  0x000000010ad0c268 in ovl_lookup (req=0x1145fca40, parent=70405, name=0x7fffbe1c0038 "toolchain-arm_xscale_gcc-7.4.0_musl_eabi") at main.c:1863         
#4  0x00007fffbe68c724 in do_lookup (req=<optimized out>, nodeid=<optimized out>, inarg=<optimized out>) at ../lib/fuse_lowlevel.c:1039                     
#5  0x00007fffbe68edfc in fuse_session_process_buf_int (se=0x113c5d390, buf=0x7fffe618bea0, ch=<optimized out>) at ../lib/fuse_lowlevel.c:2579              
#6  0x00007fffbe687bf4 in fuse_session_loop (se=0x113c5d390) at ../lib/fuse_loop.c:34                                                                       
#7  0x000000010ad03e94 in main (argc=<optimized out>, argv=<optimized out>) at main.c:5054      
```

Beyond being able to say that `pnode->children` is `NULL`, I'm at something of a loss.

giuseppe:
do you have a quick reproducer that I could use?

[...]

PR #113, opened 2019-08-29
[WIP] use st_dev for the inode key

giuseppe:
include the st_dev in the inodes map key.  Now if the same ino is used on two different layers backed by a different file system, they are treated as different inodes.

Closes: https://github.com/containers/fuse-overlayfs/issues/112
    

rhatdan:
LGTM
Makes sense.

rhatdan:
Oops I guess you marked this as a WIP, probably should not have merged.

PR #114, opened 2019-08-30
set_attrs: don't require write to S_IFREG

nwf:
It's perfectly possible to set attributes by being the file's owner; one
does not need write access to the contents.

### DIFF
--- a/main.c
+++ b/main.c
@@ -3481,7 +3481,7 @@ ovl_setattr (fuse_req_t req, fuse_ino_t ino, struct stat *attr, int to_set, stru
       switch (mode & S_IFMT)
         {
         case S_IFREG:
-          cleaned_up_fd = fd = TEMP_FAILURE_RETRY (openat (dirfd, node->path, O_NOFOLLOW|O_NONBLOCK|O_WRONLY));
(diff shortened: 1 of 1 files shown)


giuseppe:
looks like the tests are failing

giuseppe:
`ftruncate` requires the file to be open for writing

giuseppe:
@nwf are you still working on this PR?

nwf:
No, my day job has had me busy with other things.  Please feel free to adjust the patch as appropriate and close the PR.

rhatdan:
@giuseppe What do you want to do with this?

nwf:
FWIW, this is relevant because tools like tar, for example, create a file and chmod away all permissions and then attempt to modify metadata.  When that happens within a fuse-overlayfs mount, the chmod() passes through, zeroing permissions, and the subsequent modifications trigger new open()s that fail.

giuseppe:
I'll take a look at it.  It is not an issue with rootless containers, as fuse-overlayfs runs in a namespace so it still works

giuseppe:
@nwf I've opened https://github.com/containers/fuse-overlayfs/pull/120 and I've kept you as author, so please let me know if it looks fine for you before we can merge it :-)

nwf:
Replaced by #120.

Issue #116, opened 2019-08-27
buildah bud fails due to missing file system in the container


gireeshpunathil:
Failing case: missing mount binding to the parent container

```
#docker run -it --privileged quay.io/buildah/stable
[root@e8c64993a35b /]# echo 'FROM alpine' > Dockerfile
[root@e8c64993a35b /]# echo 'RUN whoami' >> Dockerfile
[root@e8c64993a35b /]# buildah bud -t foo .
STEP 1: FROM alpine
Getting image source signatures
Copying blob 9d48c3bd43c5 done
Copying config 9617696764 done
Writing manifest to image destination
Storing signatures
STEP 2: RUN whoami
process exited with error: fork/exec /bin/sh: no such file or directorysubprocess exited with status 1
                                                                                                      error building at STEP "RUN whoami": exit status 1
[root@e8c64993a35b /]# 
```

Passing case: mount binding enabled to the parent container

```
#docker run -it --privileged -v containers:/var/lib/containers quay.io/buildah/stable
[root@c4975399f13e /]# echo 'FROM alpine' > Dockerfile
[root@c4975399f13e /]# echo 'RUN whoami' >> Dockerfile
[root@c4975399f13e /]# buildah bud -t foo .
STEP 1: FROM alpine
STEP 2: RUN whoami
root
STEP 3: COMMIT foo
Getting image source signatures
Copying blob 03901b4a2ea8 skipped: already exists
Copying blob 652252567dcf done
Copying config bbf36d740b done
Writing manifest to image destination
Storing signatures
bbf36d740bc9b1defd3efb7cce0900b70f7a532863b7bc49d89c8a0c44a4d1f6
[root@c4975399f13e /]# 
```

So far so good. Now I converted this cli invocation into a tekton pipeline as follows (only the task steps are shown that is most relevant here)

```
  steps:
    - name: assemble-extract-step
      image: gireeshpunathil/foo
      command: ["/bin/bash"]
      args:
        - -c
        - "somescript.sh"
      env:
      securityContext:
        privileged: true
      volumeMounts:
        - name: varlibcontainers
          mountPath: /var/lib/containers
  volumes:
    - name: varlibcontainers
```

The image `gireeshpunathil/foo` is essentially a thin layer on top of `quay.io/buildah/stable`. 

This works as expected with `minikube`. (just like second case above)

This fails in minishift (just like first case above)

With my limited experience with `minishift`, I believe this has something to do with either:
[...]

PR #117, opened 2019-09-09
main: check whether writeback is supported

giuseppe:
disable writeback if it is not supported by FUSE.

Closes: https://github.com/containers/fuse-overlayfs/issues/116


rhatdan:
LGTM

PR #121, opened 2019-09-13
More detailed build instructions, closes #107

### LINKED ISSUE #107: Dockerfile build fails
Building the library using `builda bud` as explained in the docs 
https://github.com/containers/fuse-overlayfs/blob/f8ba9ade20d107d3e6a3dc11b6e01116894d2387/README.md#L36

fails at [this line](https://github.com/containers/fuse-overlayfs/blob/4dc60f0e157293f8692446e35d3bfc7420e36773/Dockerfile.static#L12) with error:
```
[...]
Installing /build/libfuse/doc/mount.fuse3.8 to /usr/share/man/man8
Installing /build/libfuse/build/meson-private/fuse3.pc to /usr/lib64/pkgconfig
Running custom install script '/build/libfuse/util/install_helper.sh /etc /usr/bin /usr/lib/udev/rules.d true'
mknod: /dev/fuse: Operation not permitted
FAILED: meson-install
/usr/bin/meson install --no-rebuild
ninja: build stopped: subcommand failed.
```

As far as I understand the container runtime (rightfully) prevents [this call](https://github.com/libfuse/libfuse/blob/f0e08cc700d629da2d46def8b620b0ed858cc0d9/util/install_helper.sh#L35) to `mknod`. If run in privileged mode, the call would probably succeed, but that is probably unsafe and will likely modify the host; on the other hand, this might be desired (in my case, the host already has FUSE2 installed, so it makes no difference).

Instead, I simply "tricked" the script into [believing](https://github.com/libfuse/libfuse/blob/f0e08cc700d629da2d46def8b620b0ed858cc0d9/util/install_helper.sh#L33) that the device already exists by adding the line `touch /dev/fuse` before [this one](https://github.com/containers/fuse-overlayfs/blob/4dc60f0e157293f8692446e35d3bfc7420e36773/Dockerfile.static#L12). With this workaround the build completes successfully.

Further problems I encountered:
* [this command](https://github.com/containers/fuse-overlayfs/blob/4dc60f0e157293f8692446e35d3bfc7420e36773/Dockerfile.static#L21) does not seem to work: `fuse-overlayfs: upperdir not specified`; as far as I understand, this is expected.
* it might be helpful to newcomers to explain what to do after the build (grab the binary from the image and install it somewhere in the PATH on the host).

matpen:
Just a few improvements to the `README.md` in order to facilitate newcomers.

### DIFF
--- a/README.md
+++ b/README.md
@@ -22,16 +22,50 @@ Specify a different UID/GID mapping:
 $ fuse-overlayfs -o uidmapping=0:10:100:100:10000:2000,gidmapping=0:10:100:100:10000:2000,lowerdir=lowerdir/a:lowerdir/b,upperdir=up,workdir=workdir merged
 ```
 
-Build Requirements:
+Requirements:
 =======================================================
 
-fuse-overlayfs requires libfuse > v3.2.1
+If your are not using the static build as explained in the next chapter, your system needs `libfuse` > v3.2.1.
 
-On Fedora: `dnf install fuse3-devel`
+* On Fedora: `dnf install fuse3-devel`
+* On Ubuntu > v19.04: `apt install libfuse3-dev`
+
+Also, please note that, when using `fuse-overlayfs` **from a user namespace** 
+(for example, when using rootless `podman`) a Linux Kernel > v4.18.0 is required.
 
 
 Static Build:
 =======================================================
 
-`buildah bud -t fuse-overlayfs -f ./Dockerfile.static .`
+This project provides a convenient way to automatically perform a static build using a container.
+The result is a self-contained binary without dependencies, that can be copied across hosts.
+
+1. Install `buildah` as explained [here](https://github.com/containers/buildah/blob/master/install.md);
+
+2. Both the build and deploy host require the special device `/dev/fuse`; there are a few ways to obtain it:
+
+  * install `fuse2` or `fuse3` using the package manager of your choice (dnf, apt, pacman, etc): 
+the install script will usually create the device automatically; or
+  * manually create the device with the command `mknod /dev/fuse -m 0666 c 10 229` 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -22,16 +22,49 @@ Specify a different UID/GID mapping:
 $ fuse-overlayfs -o uidmapping=0:10:100:100:10000:2000,gidmapping=0:10:100:100:10000:2000,lowerdir=lowerdir/a:lowerdir/b,upperdir=up,workdir=workdir merged
 ```
 
-Build Requirements:
+Requirements:
 =======================================================
 
-fuse-overlayfs requires libfuse > v3.2.1
+Please note that `fuse-overlayfs` requires Linux Kernel > v4.8.0.
```
giuseppe: why is Linux v4.8.0 required?

caemate: I faced this problem myself at the time of first install, and finally spotted [your comment](https://github.com/containers/libpod/issues/3109#issuecomment-491848145) that clarifies this.

In addition, there is a typo in my code, which I now fixed!

matpen: I faced this problem myself at the time of first install, and finally spotted [your comment](https://github.com/containers/libpod/issues/3109#issuecomment-491848145) that clarifies this.

In addition, there is a typo in my code, which I now fixed!

giuseppe:
I see thanks!  Could you drop that part completely or at least specify (to work in a user namespace)?

Older versions of Linux are still fine if you don't use fuse-overlayfs from a user namespace (such as rootless Podman).

Nit, could you squash the patches together?

matpen:
Done! I tried to rephrase to include that precious information. Let me know your thoughts, and I can refine if necessary.

> Nit, could you squash the patches together?

Sure, it should look like one clean change now.

giuseppe:
Thanks, LGTM!

PR #123, opened 2019-09-23
main: fix skip whiteout deletion optimization

giuseppe:
fix the lookup function to report also a whiteout node, instead of NULL.  This is required to correctly delete the whiteout file when it exists.

Closes: https://github.com/containers/fuse-overlayfs/issues/122


giuseppe:
@rhatdan PTAL

rhatdan:
LGTM

Issue #124, opened 2019-09-29
Failure to invalidate cache on chmod


nwf:
On the host, we begin:

```
veloci% ls -l /dozer/scratch/openwrt/images/darknetgw/upper/tmp/stage-perl/usr/lib/perl5/5.28/CORE/regexp.h
-r--r--r-- 1 nwf nwf 36363 Sep 29 18:24 /dozer/scratch/openwrt/images/darknetgw/upper/tmp/stage-perl/usr/lib/perl5/5.28/CORE/regexp.h  
```

Inside the overlayfs mount, inside an unprivileged user container, we run:

```
bash-5.0$ ls -l '/opt/build/tmp/stage-perl/usr/lib/perl5/5.28/CORE/uconfig.h'
-r--r--r-- 1 daemon 1 162891 Sep 29 18:23 /opt/build/tmp/stage-perl/usr/lib/perl5/5.28/CORE/uconfig.h
bash-5.0$ chmod 644 '/opt/build/tmp/stage-perl/usr/lib/perl5/5.28/CORE/regexp.h'
bash-5.0$ ls -l '/opt/build/tmp/stage-perl/usr/lib/perl5/5.28/CORE/uconfig.h'
-r--r--r-- 1 daemon 1 162891 Sep 29 18:23 /opt/build/tmp/stage-perl/usr/lib/perl5/5.28/CORE/uconfig.h
```

However, on the host, the file actually has been modified:

```
veloci% ls -l /dozer/scratch/openwrt/images/darknetgw/upper/tmp/stage-perl/usr/lib/perl5/5.28/CORE/regexp.h
-rw-r--r-- 1 nwf nwf 36363 Sep 29 18:24 /dozer/scratch/openwrt/images/darknetgw/upper/tmp/stage-perl/usr/lib/perl5/5.28/CORE/regexp.h
```

nwf:
(This was tested on b0a9bdaa646ac4bcd6e68cc6489353d52d157af7; I don't believe any of the subsequent changes are relevant.)

giuseppe:
I am not able to see it locally, so I am not sure about the fix.

Could you please try if https://github.com/containers/fuse-overlayfs/pull/126 fixes the issue for you?

PR #126, opened 2019-10-02
main: store the new mode after chmod

giuseppe:
after a successful chmod, store the new mode for the ino.

Closes: https://github.com/containers/fuse-overlayfs/issues/124


giuseppe:
let's iterate more on te issue if required, but I think this fix is necessary anyway, so let's merge

rhatdan:
LGTM

PR #128, opened 2019-10-03
writeback: restore suid mode after write

giuseppe:
restore the inode mode if there is a writepage request that could
reset the setuid bit.

Closes: https://github.com/containers/fuse-overlayfs/issues/127


giuseppe:
@rhatdan there are a few open PRs, I'll cut a new release once we get them merged

PR #129, opened 2019-10-03
main: skip whiteout creation if not needed

giuseppe:
if the file is present only on the upper layer, do not create a
whiteout when it is created.  Previously it was looking only at the
parent directory.


rhatdan:
LGTM

PR #132, opened 2019-10-15
main.c: reload the directory on stat

giuseppe:
make sure the directory is reloaded so the correct number of links can
be retrieved.

Closes: https://github.com/containers/fuse-overlayfs/issues/131


rhatdan:
LGTM

Issue #100, opened 2019-08-07
fuse-overlay needs to support readonly lower with no upper.


rhatdan:
Containers storage, has a fix that allows additional-stores to be mounted read/only, but this requires the mounting of multi-layer lowers without an upper directory.  Currently fuse-overlay does not support this and is has to be disabled from this feature.

rhatdan:
@nalind FYI

rhatdan:
https://github.com/containers/storage/blob/master/drivers/overlay/overlay.go#L844

AkihiroSuda:
@giuseppe Is there any specific reason not to support read-only mode?

I'm trying to implement containerd snapshotter for fuse-overlayfs and want read-only mode

AkihiroSuda:
These parts require read-only multi-lowerdir mount

https://github.com/containerd/containerd/blob/36c4260e6fa15cb20012cfe95f322ef85cf771ca/snapshots/overlay/overlay.go#L469-L501

giuseppe:
> @giuseppe Is there any specific reason not to support read-only mode?

no, only I had not time to look into it yet.  It should not be too difficult as it is only a matter of checking fuse-overlayfs returns `EROFS` on each operation that attempts to modify the file system.  I'll take a look

giuseppe:
I've opened a PR here: https://github.com/containers/fuse-overlayfs/pull/133

PR #133, opened 2019-10-19
fuse-overlayfs: support running without upper layer

giuseppe:
when there is no upper layer specified, fail every read operation with
EROFS.

Closes: https://github.com/containers/fuse-overlayfs/issues/100

### DIFF
--- a/README.md
+++ b/README.md
@@ -3,12 +3,6 @@ fuse-overlayfs
 
 An implementation of overlay+shiftfs in FUSE for rootless containers.
 
-Limitations:
-=======================================================
-
-Read-only mode is not supported, so it is always required to specify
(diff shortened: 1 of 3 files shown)


giuseppe:
> thanks, but README needs update

thanks, fixed now

AkihiroSuda:
FYI I'll use this in https://github.com/AkihiroSuda/containerd-fuse-overlayfs

rhatdan:
@AkihiroSuda Could something like that be done for CRI-O.

AkihiroSuda:
CRI-O is using `containers/storage` which already supports fuse-overlayfs?

rhatdan:
Ok, I was not quite understanding what your tool was doing.  Great.

Issue #143, opened 2019-11-15
error setting attributes on dangling symlink


nalind:
I think I've got https://github.com/containers/buildah/issues/1979 narrowed down to this reproducer:
```
#!/bin/bash
mount="mount -t overlay"
mount=fuse-overlayfs
mkdir lower/ upper/ work/ merged/
touch lower/deps.txt
ln -s src/deps.txt upper/
$mount -o lowerdir=`pwd`/lower,upperdir=`pwd`/upper,workdir=`pwd`/work none `pwd`/merged
touch -h merged/deps.txt
```

With `mount=fuse-overlayfs`, the `touch` command produces an ELOOP error.  Possibly a symptom of the same bug: if we don't create a `deps.txt` in the lower directory, `touch` appears to succeed, but the timestamps aren't changed.

With `mount="mount -t overlay"`, the timestamp on the symlink is successfully updated either way.

giuseppe:
It seems the issue is fixed with 62abdcf00f8dfa5aff3fcfad9164dd7365589353

https://bodhi.fedoraproject.org/updates/FEDORA-2019-b71b08d561

I've used your reproducer to add a new test in fuse-overlayfs and opened a PR (patch in your name): https://github.com/containers/fuse-overlayfs/pull/144

Could you please check if it looks fine for you?

PR #144, opened 2019-11-15
tests: test that touch -h works

giuseppe:
verify that setting attributes on a dangling symlink work, even if the
file type is different on a lower layer.

Closes: https://github.com/containers/fuse-overlayfs/issues/143


rhatdan:
LGTM

PR #146, opened 2019-11-18
configure: check for FUSE copy_file_range

giuseppe:
check if the version of FUSE supports copy_file_range.

It fixes the build on RHEL 8.


rhatdan:
LGTM

Issue #151, opened 2019-11-27
Whiteout file in root directory globally causes all files of same name to appear opaque


eddsalkield:
If one creates file `/.wh.test` in an upper layer, then for any subdirectory, all files named `test` are opaque.

This is reproducible as follows:
Within the layer on the host filesystem, create in the root directory a file named `.wh.test`

Now mount the _fuse-overlayfs_.  Then within the mounted filesystem, run:
```
$ touch /root/test
$ file /root/test
/root/test: empty
```
After a remount:
```
$ file /root/test
test: cannot open `test' (No such file or directory)
```

giuseppe:
@eddsalkield thanks for the report.  I've opened a PR to address it: https://github.com/containers/fuse-overlayfs/pull/152

Could you please verify if it solves the issue for you as well?

eddsalkield:
This solves the issue for me - thanks very much!

PR #152, opened 2019-11-27
main: prevent parent whiteout to hide file

giuseppe:
use the full relative path when looking up the whiteout file,
otherwise a whiteout in the upper layer will hide files in
subdirectories.

Closes: https://github.com/containers/fuse-overlayfs/issues/151


rhatdan:
LGTM

PR #156, opened 2019-12-06
Simpler static build

leorochael:
I had trouble compiling the static binary in Ubuntu 18.04 using the `buildah` instructions in the README.

The currently available `buildah bud` in `buildah` version (1.10.1) doesn't have a `--device` switch, so I couldn't find a way to expose the requested `/dev/fuse` device into the container which broke the internal `libfuse` build.

Luckily, more recent ubuntu versions already have `libfuse3-dev` with a static library, so the whole `libfuse` build can be skipped.

As a side-effect the whole build process is a lot faster, and using the `--volume` switch to `buildah bud` makes the build instructions simpler for copying the built binary into the deployment host.

So I'd like to offer this PR. Feel free to suggest changes or improvements.

In particular, I could replace `Dockerfile.static` with one based on `Dockerfile.simpler` instead of keeping two `Dockerfile`s around.


Dockerfile.simpler
```diff
@@ -0,0 +1,11 @@
+FROM docker.io/ubuntu:rolling
+WORKDIR /build
+RUN apt-get update && \
+    apt-get install --no-install-recommends -y \ &&
```
giuseppe: it looks like the `&&` is after the `\`.  It should be moved to the new line

leorochael: there should be no `&&` at all here, actually.

leorochael:
> I think we can keep both versions, maybe we rename them `Dockerfile.static.fedora` and `Dockerfile.static.ubuntu`? What do you think?

I could do that.

What about the instructions in the REAME? Should they point to `Dockerfile.static.ubuntu` for static builds then?

Also, is `Dockerfile.static` used for anything beside the static build of `fuse-overlayfs`? Specifically, is the `ENTRYPOINT ["/usr/bin/fuse-overlayfs","-f"]` command actually useful? Should I replicate it in `Dockerfile.static.ubuntu`?

Finally, I'm getting some warnings (in the link below) while compiling `main.c`. Are they expected? Or are they a result of compiling fuse-overlayfs in ubuntu (19.10, in this case)?

[build-warnings.txt](https://github.com/containers/fuse-overlayfs/files/3933830/build-warnings.txt)

giuseppe:
> Also, is `Dockerfile.static` used for anything beside the static build of `fuse-overlayfs`? Specifically, is the `ENTRYPOINT ["/usr/bin/fuse-overlayfs","-f"]` command actually useful? Should I replicate it in `Dockerfile.static.ubuntu`?

the original author was @vbatts.

@vbatts, were you using it to run fuse-overlayfs inside of a container?

I don't think it needs to be part of the entrypoint, so your Dockerfile looks correct as it is.

> What about the instructions in the REAME? Should they point to `Dockerfile.static.ubuntu` for static builds then?

yes please

vbatts:
I don't recall the use-case currently ☹

leorochael:
I've renamed the `static` dockerfiles as suggested.

If you have no further requests, please feel free to merge it.

giuseppe:
thanks, LGTM!

PR #163, opened 2020-01-08
Fix possible segmentation fault in direct_fsync()

saschagrunert:
If the call to `get_upper_layer(lo)` returns `NULL` then the
`node->layer` will be `NULL`, too. If this is the case we pass `NULL` to
`direct_fsync()` which will cause a segmentation fault in:

```c
cfd = openat(l->fd, path, O_NOFOLLOW|O_DIRECTORY);
```

To fix this we now apply an additional check and error in the case of
`get_upper_layer(lo) == NULL`.

### DIFF
--- a/main.c
+++ b/main.c
@@ -4595,6 +4595,12 @@ do_fsync (fuse_req_t req, fuse_ino_t ino, int datasync, int fd)
   /* Skip fsync for lower layers.  */
   do_fsync = node && node->layer == get_upper_layer (lo);
 
+  if (node->layer == NULL)
+    {
+      fuse_reply_err (req, ENOENT);
+      return;
+    }
+
   if (fd < 0)
     strcpy (path, node->path);
 


giuseppe:
I am trying to figure out how this would happen but I don't see in `do_fsync` any assignment to `node->layer`.  Does it happen in another function?

saschagrunert:
> I am trying to figure out how this would happen but I don't see in `do_fsync` any assignment to `node->layer`. Does it happen in another function?

No the access occurs in `direct_fsync` here:
https://github.com/containers/fuse-overlayfs/blob/0aa053745904c0c842c8ac7a81f804839406a29f/main.c#L4561

I just put the error handling a into `do_fsync` to have a unique place for those.

giuseppe:
the extra check won't hurt!  Thanks for working on it.

Just out of curiosity, have you found it with some static analysis tool or just checking the code?

saschagrunert:
> the extra check won't hurt! Thanks for working on it.
> 
> Just out of curiosity, have you found it with some static analysis tool or just checking the code?

Both, :) but I could verify my assumption with [scan-build](https://clang-analyzer.llvm.org/scan-build.html). Unfortunately it produces also false-positives in relation to the cleanup attributes.

Issue #160, opened 2020-01-07
Files not removalable/masked 


phonedph1:
I'm not sure if this is a bug or if I'm doing something wrong here, but I seem to be able to get into a state where some files are not removable, and if you overwrite and then delete them the original re-appears.

This can be reproduced on:
```
$ fuse-overlayfs -V
fuse-overlayfs: version 0.7.2
FUSE library version 3.9.0
using FUSE kernel interface version 7.31
```

using CentOS 8.1 with the shipped podman `podman-1.4.2-5.module_el8.1.0+237+63e26edc.x86_64`.

Here is how I can reproduce it:

```
$ cat Dockerfile
FROM centos:8
RUN dnf -y install vim
RUN dnf -y install tmux
RUN dnf -y clean all
```

And then:

```
$ podman run -it hrm /bin/bash
[root@a09d26610a61 /]# dnf -y update bash
Failed to set locale, defaulting to C
CentOS-8 - AppStream                                                                                                                                                               1.3 MB/s | 6.3 MB     00:04
CentOS-8 - Base                                                                                                                                                                    1.7 MB/s | 7.9 MB     00:04
CentOS-8 - Extras                                                                                                                                                                  662  B/s | 2.1 kB     00:03
Dependencies resolved.
===================================================================================================================================================================================================================
 Package                                        Arch                                             Version                                                    Repository                                        Size
===================================================================================================================================================================================================================
Upgrading:
 bash                                           x86_64                                           4.4.19-8.el8_0                                             BaseOS                                           1.5 M

Transaction Summary
===================================================================================================================================================================================================================
Upgrade  1 Package

Total download size: 1.5 M
Downloading Packages:
bash-4.4.19-8.el8_0.x86_64.rpm                                                                                                                                                     459 kB/s | 1.5 MB     00:03
-------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
Total                                                                                                                                                                              326 kB/s | 1.5 MB     00:04
Running transaction check
Transaction check succeeded.
Running transaction test
Transaction test succeeded.
Running transaction
  Preparing        :                                                                                                                                                                                           1/1
  Upgrading        : bash-4.4.19-8.el8_0.x86_64                                                                                                                                                                1/2
  Running scriptlet: bash-4.4.19-8.el8_0.x86_64                                                                                                                                                                1/2
  Cleanup          : bash-4.4.19-7.el8.x86_64                                                                                                                                                                  2/2
[...]

PR #164, opened 2020-01-08
main: force a dir reload after a move

giuseppe:
if the directory was moved, invalidate its cache and force a reload.

Closes: https://github.com/containers/fuse-overlayfs/issues/160


rhatdan:
LGTM

giuseppe:
@phonedph1 confirmed the patch solves the issue

Issue #167, opened 2020-01-13
Error running MSSQL image


abitrolly:
/kind bug
**Description**

I needed to run [this image](https://hub.docker.com/_/microsoft-mssql-server?tab=description) to [port this app](https://github.com/epam/OSCI/issues/2) to another DB.  Using officially provided `docker` command didn't work.

<details>
  <summary>Click to expand</summary>
<pre>
✗ podman run -e 'ACCEPT_EULA=Y' -e 'SA_PASSWORD=yourStrong(!)Password' -p 1433:1433 -it mcr.microsoft.com/mssql/server:2017-CU8-ubuntu 
Trying to pull mcr.microsoft.com/mssql/server:2017-CU8-ubuntu...
Getting image source signatures
Copying blob 5594e4e5950b done  
Copying blob b70745c852a2 done  
Copying blob 4fa80d7b805d done  
Copying blob 484dd0f2fbdc done  
Copying blob 718060832ef2 done  
Copying blob 47004b22ec62 done  
Copying blob 7d648891de3f done  
Copying blob 5b67719e2956 done  
Copying blob e0d1b3db20c8 done  
Copying blob ded313a21911 done  
Copying config 229d30f7b4 done  
Writing manifest to image destination
Storing signatures
^[	2020-01-13 14:08:10.45 Server      Setup step is copying system data file 'C:\templatedata\master.mdf' to '/var/opt/mssql/data/master.mdf'.
2020-01-13 14:08:10.55 Server      Did not find an existing master data file /var/opt/mssql/data/master.mdf, copying the missing default master and other system database files. If you have moved the database location, but not moved the database files, startup may fail. To repair: shutdown SQL Server, move the master database to configured location, and restart.
2020-01-13 14:08:10.55 Server      Setup step is copying system data file 'C:\templatedata\mastlog.ldf' to '/var/opt/mssql/data/mastlog.ldf'.
2020-01-13 14:08:10.56 Server      Setup step is copying system data file 'C:\templatedata\model.mdf' to '/var/opt/mssql/data/model.mdf'.
2020-01-13 14:08:10.59 Server      Setup step is copying system data file 'C:\templatedata\modellog.ldf' to '/var/opt/mssql/data/modellog.ldf'.
2020-01-13 14:08:10.61 Server      Setup step is copying system data file 'C:\templatedata\msdbdata.mdf' to '/var/opt/mssql/data/msdbdata.mdf'.
[...]

PR #168, opened 2020-01-14
main: reset O_DIRECT on open

giuseppe:
having the fd open with O_DIRECT causes pwrite64 to fail with EINVAL if the write buffer is not aligned.  Since we have no control on the buffer provided by libfuse, just drop O_DIRECT.
    
Closes: https://github.com/containers/fuse-overlayfs/issues/167
    

rhatdan:
LGTM

Issue #183, opened 2020-02-13
groupadd fails inside container when run rootless


castedo:
/kind bug

**Description**

Executing "groupadd foo" within a debian or ubuntu container fails when run by rootless podman but works when run by root/sudo podman (podman run from RHEL8). I'm not sure if this is a bug or an indication https://github.com/containers/libpod/blob/master/rootless.md should get updated and maybe an enhancement should be made. Either way, I found this behavior surprising and confusing.

The impact of this is that many debian/ubuntu packages will fail to get-apt install when run inside a rootless container because many packages will have some dependency that will run groupadd within an install script (e.g. dbus package).

In addition to "debian:10.2" I also repro with more recent "debian:bullseye-20200130" and "ubuntu:19.10" and "ubuntu:focal-20200115".

**Steps to reproduce the issue:**

1. as non-root run `podman run -it --rm library/debian:10.2`

2. inside container run `groupadd foo`

**Describe the results you received:**

```
groupadd: /etc/group.7: lock file already used
groupadd: cannot lock /etc/group; try again later.
```

**Describe the results you expected:**

Expected it to succeed with no output which is what happens when running the same podman command but as root/sudo.

**Output of `podman version`:**

```
podman version 1.6.4
```

**Output of `podman info --debug`:**

```
debug:
  compiler: gc
  git commit: ""
  go version: go1.12.12
  podman version: 1.6.4
host:
  BuildahVersion: 1.12.0-dev
  CgroupVersion: v1
  Conmon:
    package: conmon-2.0.6-1.module+el8.1.1+5259+bcdd613a.x86_64
    path: /usr/bin/conmon
    version: 'conmon version 2.0.6, commit: 6ffbb2ec70dbe5ba56e4bfde946fb04f19dd8bbf'
  Distribution:
    distribution: '"rhel"'
    version: "8.1"
  IDMappings:
    gidmap:
    - container_id: 0
      host_id: 1000
      size: 1
    - container_id: 1
      host_id: 100000
      size: 65536
    uidmap:
    - container_id: 0
      host_id: 1000
      size: 1
    - container_id: 1
      host_id: 100000
      size: 65536
  MemFree: 472834048
  MemTotal: 7902752768
  OCIRuntime:
    name: runc
    package: runc-1.0.0-64.rc9.module+el8.1.1+5259+bcdd613a.x86_64
    path: /usr/bin/runc
    version: 'runc version spec: 1.0.1-dev'
  SwapFree: 4294176768
[...]

PR #184, opened 2020-02-15
main: force timeout 0 for ovl_link

giuseppe:
There is an issue on RHEL 8.1 where the nlink counter is always
incremented by one, no matter what is specified in e.attr.st_nlink.

Always set timeout to 0 to force a new stat on the inode.

Closes: https://bugzilla.redhat.com/show_bug.cgi?id=1802907
Closes: https://github.com/containers/fuse-overlayfs/issues/183


giuseppe:
@szmi PTAL

rhatdan:
LGTM

rhatdan:
@szmi  does this fix the issue for you?

giuseppe:
I've pinged @szmi as the issue seems to be in FUSE on RHEL 8.1 rather than fuse-overlayfs.

This is only a workaround.

giuseppe:
I think we can merge this PR

giuseppe:
it is going to be addressed in the kernel, so closing it

Issue #185, opened 2020-02-18
help message is outdated


progrunner17:
According to the help message, `-d` option or  ` -o debug` option should enable the debug output, 
but it won't work while `--debug` option makes it.

I found the defect with my self-build fuse-overlayfs (commmit 8bba496f08787344ab45b36b0640d50b85edab78 ).

I'm sorry if the problem has already been fixed on the other commit or PR.

rhatdan:
@progrunner17 interested in opening a PR to fix the help message?

rhatdan:
I opened a PR.

PR #186, opened 2020-02-18
Add support for debug to match help

rhatdan:
The help message that is printed does not match reality for the command line.

Signed-off-by: Daniel J Walsh <dwalsh@redhat.com>


rhatdan:
Fixes: https://github.com/containers/fuse-overlayfs/issues/185

Issue #189, opened 2020-02-28
Ghost directory in rootless podman software build


michaelweiser:
I am trying to build openembedded-core in a rootless podman container running on fuse-overlayfs. It works fine with storage vfs but reproducibly fails at the same point with fuse-overlayfs. The error I'm getting is:
```
Currently  4 running tasks (328 of 1409)  23% |######################                                                                             |
ERROR: linux-libc-headers-5.2-r0 do_package: Error executing a python function in exec_python_func() autogenerated:

The stack trace of python calls that resulted in this exception/failure was:
'exec_python_func() autogenerated', lineno: 2, function: <module>
     0001:
 *** 0002:sstate_task_postfunc(d)
     0003:
'/home/bitbake/openembedded-core-2019-10.2-zeus/meta/classes/sstate.bbclass', lineno: 732, function: sstate_task_postfunc
     0728:
     0729:    omask = os.umask(0o002)
     0730:    if omask != 0o002:
     0731:       bb.note("Using umask 0o002 (not %0o) for sstate packaging" % omask)
 *** 0732:    sstate_package(shared_state, d)
     0733:    os.umask(omask)
     0734:
     0735:    sstateinst = d.getVar("SSTATE_INSTDIR")
     0736:    d.setVar('SSTATE_FIXMEDIR', shared_state['fixmedir'])
'/home/bitbake/openembedded-core-2019-10.2-zeus/meta/classes/sstate.bbclass', lineno: 647, function: sstate_package
     0643:        if sharedworkdir in plain:
     0644:            pdir = plain.replace(sharedworkdir, sstatebuild)
     0645:        bb.utils.mkdirhier(plain)
     0646:        bb.utils.mkdirhier(pdir)
 *** 0647:        os.rename(plain, pdir)
     0648:
     0649:    d.setVar('SSTATE_BUILDDIR', sstatebuild)
     0650:    d.setVar('SSTATE_INSTDIR', sstatebuild)
     0651:
Exception: OSError: [Errno 39] Directory not empty: '/home/bitbake/openembedded-core-2019-10.2-zeus/build/tmp-glibc/work/armv5b-oe-linux-gnueabi/linux-libc-headers/5.2-r0/package' -> '/home/bitbake/openembedded-core-2019-10.2-zeus/build/tmp-glibc/work/armv5b-oe-linux-gnueabi/linux-libc-headers/5.2-r0/sstate-build-package//package'

[...]

PR #191, opened 2020-03-03
main, rename: make sure destination is deleted

giuseppe:
before doing a rename(2) make sure the destination is cleaned up,
otherwise a left-over directory can cause the rename to fail with
EEXIST.

Closes: https://github.com/containers/fuse-overlayfs/issues/189


rhatdan:
LGTM

PR #193, opened 2020-03-15
Set st_nlink correctly when stat is called more than once for the same path

dhedberg:
The directory link count reported by ls is correct the first time, but is then always set to 1 if you run it again.

### DIFF
--- a/main.c
+++ b/main.c
@@ -701,18 +701,23 @@ rpl_stat (fuse_req_t req, struct ovl_node *node, int fd, const char *path, struc
 
   st->st_ino = node->tmp_ino;
   st->st_dev = node->tmp_dev;
-  if (ret == 0 && node_dirp (node) && node->ino->nlinks <= 0)
+  if (ret == 0 && node_dirp (node))
     {
-      struct ovl_node *it;
-
-      st->st_nlink = 2;
-
-      for (it = hash_get_first (node->children); it; it = hash_get_next (node->children, it))
+      if (node->ino->nlinks <= 0)
         {
-          if (node_dirp (it))
-            st->st_nlink++;
+          struct ovl_node *it;
+
+          st->st_nlink = 2;
+
+          for (it = hash_get_first (node->children); it; it = hash_get_next (node->children, it))
+            {
+              if (node_dirp (it))
+                st->st_nlink++;
+            }
+          node->ino->nlinks = st->st_nlink;
         }
-      node->ino->nlinks = st->st_nlink;
+      else
+        st->st_nlink = node->ino->nlinks;
     }
 
   return ret;


giuseppe:
thanks for the patch, I see there is still an issue after `rmdir` and the nlink count is not correct.

I think we should better take rid of this optimization, could you try if the following patch works for you?

```
diff --git a/fuse-overlayfs.h b/fuse-overlayfs.h
index 26a74b5..ed810f8 100644
--- a/fuse-overlayfs.h
+++ b/fuse-overlayfs.h
@@ -33,7 +33,6 @@ struct ovl_ino
   dev_t dev;
   int lookups;
   mode_t mode;
-  int nlinks;
 };
 
 struct ovl_node
diff --git a/main.c b/main.c
index 5b00d80..f14af99 100644
--- a/main.c
+++ b/main.c
@@ -690,8 +690,8 @@ rpl_stat (fuse_req_t req, struct ovl_node *node, int fd, const char *path, struc
     ret = stat (path, st);
   else if (node->hidden)
     ret = fstatat (node_dirfd (node), node->path, st, AT_SYMLINK_NOFOLLOW);
-    else
-      ret = l->ds->statat (l, node->path, st, AT_SYMLINK_NOFOLLOW, STATX_BASIC_STATS);
+  else
+    ret = l->ds->statat (l, node->path, st, AT_SYMLINK_NOFOLLOW, STATX_BASIC_STATS);
 
   if (ret < 0)
     return ret;
@@ -701,7 +701,7 @@ rpl_stat (fuse_req_t req, struct ovl_node *node, int fd, const char *path, struc
 
   st->st_ino = node->tmp_ino;
   st->st_dev = node->tmp_dev;
-  if (ret == 0 && node_dirp (node) && node->ino->nlinks <= 0)
+  if (ret == 0 && node_dirp (node))
     {
       struct ovl_node *it;
 
@@ -712,7 +712,6 @@ rpl_stat (fuse_req_t req, struct ovl_node *node, int fd, const char *path, struc
           if (node_dirp (it))
             st->st_nlink++;
         }
-      node->ino->nlinks = st->st_nlink;
     }
 
   return ret;
@@ -1347,19 +1346,13 @@ insert_node (struct ovl_node *parent, struct ovl_node *item, bool replace)
     {
       if (hash_lookup (prev_parent->children, item) == item)
         hash_delete (prev_parent->children, item);
-      if (is_dir)
-        prev_parent->ino->nlinks--;
     }
 
   if (replace)
     {
       old = hash_delete (parent->children, item);
       if (old)
-        {
-          if (node_dirp (old))
-            parent->ino->nlinks--;
-          node_free (old);
-        }
+        node_free (old);
     }
 
   ret = hash_insert_if_absent (parent->children, item, (const void **) &old);
@@ -1376,8 +1369,6 @@ insert_node (struct ovl_node *parent, struct ovl_node *item, bool replace)
     }
 
   item->parent = parent;
-  if (is_dir)
-    parent->ino->nlinks++;
 
   return item;
 }
```

If it works, please feel free to amend it in your PR

dhedberg:
Thanks, although I was somewhat hoping to perhaps nudge the development in the opposite direction :-)

Looking up all children for all parent directories when running stat on a file is prohibitively expensive when you have an NFS mount with many files as a lower directory. One (perhaps too poorly researched) option would be to instead base it on the actual stat from the lower filesystem and subtract whiteouts from the upper directories. I guess some kind of internal caching of the count would be necessary in that case.

I don't know if there are any potential problems with that solution, if you have any better ideas, or if this is not a concern for this project.

(I naively suspect that just hard coding nlink to 2 for all directories would work fairly well for my specific use case, but I might be missing something)

giuseppe:
> Looking up all children for all parent directories when running stat on a file is prohibitively expensive when you have an NFS mount with many files as a lower directory. One (perhaps too poorly researched) option would be to instead base it on the actual stat from the lower filesystem and subtract whiteouts from the upper directories. I guess some kind of internal caching of the count would be necessary in that case.

the main issue with such approach is that there can be multiple lower layers, and each of them can hide directories from the lower layers, so in the end we'd still need to look up each one of them.

This code should not really be executed for each stat, as the kernel itself is caching the result.

Do you have any numbers on what improvements do we have with the optimization?

dhedberg:
> This code should not really be executed for each stat, as the kernel itself is caching the result.

> Do you have any numbers on what improvements do we have with the optimization?

I don't mean that the loop over the hash table here is expensive, I haven't tested the performance but your patch is probably very reasonable. I can update my pull request later, but feel free to close this and apply your patch directly if that is faster or more convenient.

One thing we do is to basically run stat on a plain file with a known path that happens to reside in a directory containing about 90k files. Due to the call to load_dir in ovl_lookup, this takes over 1 minute even though none of the data it loads is actually useful for the stat call. Running stat directly against the nfs mount is basically instantaneous.

Since our mounts are pretty short-lived (part of a testsuite), caching is not of much help. There are probably also other scenarios where calls to load_dir is a performance problem in this case, so I was thinking that the "correct" fix might have been to keep the internal nlinks and base it on stat instead.

Judging by your answer, my use case and the goals of the project might simply be incompatible.

giuseppe:
> > 
> 
> I don't mean that the loop over the hash table here is expensive, I haven't tested the performance but your patch is probably very reasonable. I can update my pull request later, but feel free to close this and apply your patch directly if that is faster or more convenient.

yes please, and I'd like to know the issue you've seen is really fixed by it :-)

We can think of ways of improving your use case, perhaps we could add an option to fuse-overlayfs to not try to reload the directory on a stat and report a static nlink, if the user wishes so.  Would that help?

giuseppe:
alternative PR here: https://github.com/containers/fuse-overlayfs/pull/194

rhatdan:
I merged the other patch. @dhedberg Does this fix the issue for you?  If yes then we can close this PR.
@dhedberg Thanks for your help.

dhedberg:
I have verified that the issue is fixed by #194. Thank you.

> We can think of ways of improving your use case, perhaps we could add an option to fuse-overlayfs to not try to reload the directory on a stat and report a static nlink, if the user wishes so. Would that help?

I think that such an option would work great for us, but I will have to do some testing to be sure.

PR #194, opened 2020-03-16
main: drop nlink optimization

giuseppe:
calculate the nlink for the directory on each stat.


rhatdan:
LGTM

PR #195, opened 2020-03-16
main: stop ino lookup without an open fd

giuseppe:
stop the lookup for the dev/ino if there is not an ino open as it also
prevents reading the origin xattr.


rhatdan:
LGTM

Issue #197, opened 2020-03-30
Can't run podman container: open `proc`: no such file or directory: oci runtime command not found error


nightuser:
OS: ArchLinux armv7h
fuse-overlayfs 0.7.8
podman 1.8.2

With fuse-overlayfs driver I get the following error when creating a podman container:

```
$ podman run --log-level debug ubuntu
DEBU[0000] Reading configuration file "/home/nightuser/.config/containers/libpod.conf" 
DEBU[0000] Using conmon: "/usr/bin/conmon"              
DEBU[0000] Initializing boltdb state at /home/nightuser/.local/share/containers/storage/libpod/bolt_state.db 
DEBU[0000] Using graph driver overlay                   
DEBU[0000] Using graph root /home/nightuser/.local/share/containers/storage 
DEBU[0000] Using run root /run/user/1000/containers     
DEBU[0000] Using static dir /home/nightuser/.local/share/containers/storage/libpod 
DEBU[0000] Using tmp dir /run/user/1000/libpod/tmp      
DEBU[0000] Using volume path /home/nightuser/.local/share/containers/storage/volumes 
DEBU[0000] Set libpod namespace to ""                   
DEBU[0000] No store required. Not opening container store. 
DEBU[0000] Initializing event backend journald          
DEBU[0000] using runtime "/usr/bin/crun"                
DEBU[0000] using runtime "/usr/bin/runc"                
DEBU[0000] using runtime "/usr/bin/crun"                
INFO[0000] running as rootless                          
DEBU[0000] Reading configuration file "/home/nightuser/.config/containers/libpod.conf" 
DEBU[0000] Using conmon: "/usr/bin/conmon"              
DEBU[0000] Initializing boltdb state at /home/nightuser/.local/share/containers/storage/libpod/bolt_state.db 
DEBU[0000] Using graph driver overlay                   
DEBU[0000] Using graph root /home/nightuser/.local/share/containers/storage 
DEBU[0000] Using run root /run/user/1000/containers     
DEBU[0000] Using static dir /home/nightuser/.local/share/containers/storage/libpod 
DEBU[0000] Using tmp dir /run/user/1000/libpod/tmp      
DEBU[0000] Using volume path /home/nightuser/.local/share/containers/storage/volumes 
DEBU[0000] Set libpod namespace to ""                   
DEBU[0000] [graphdriver] trying provided driver "overlay" 
DEBU[0000] overlay: mount_program=/usr/bin/fuse-overlayfs 
[...]

PR #199, opened 2020-04-15
fuse-overlayfs: add missing definitions for _FILE_OFFSET_BITS

giuseppe:
it caused an issue on armv7h where different versions of dirent struct
were used in main.c and in the other files.

Regression introduced with c2c2ac5b82fb59322da227d196214b4a58ede634

Closes: https://github.com/containers/fuse-overlayfs/issues/197


rhatdan:
LGTM

PR #203, opened 2020-04-21
main: use openat2 where available

giuseppe:


rhatdan:
LGTM

srd424:
Looks like this may change in systemd if https://github.com/systemd/systemd/pull/18342 gets merged..

srd424:
systemd/systemd#18342 now merged and will presumably ship in v248; as the original check only landed in 247, it's probably not worth worrying too much about nasty workarounds in overlayfs?

giuseppe:
> [systemd/systemd#18342](https://github.com/systemd/systemd/pull/18342) now merged and will presumably ship in v248; as the original check only landed in 247, it's probably not worth worrying too much about nasty workarounds in overlayfs?

yes I agree, if the system reports EPERM that must be handled as such.  ENOSYS is the correct error message in this case and it is good that it is fixed in systemd

Issue #204, opened 2020-04-24
touch -h fails to set the date for symlinks


progandy:
I was experimenting with reproducible builds in rootless containers and found fuse-overlayfs is unable to change the modification time of symlinks which is needed to set uniform timestamps to all files in the package. The kernel module does it without problems.

```bash
#!/bin/bash
# test-symlink-touch.sh 
# sudo test-symlink-touch.sh kernel
if [[ $1 == kernel ]]; then
mount="mount -t overlay"
umount=umount
else
mount=fuse-overlayfs
umount="fusermount -u"
fi
mkdir lower/ upper/ work/ merged/
$mount -o lowerdir=`pwd`/lower,upperdir=`pwd`/upper,workdir=`pwd`/work none `pwd`/merged
touch merged/afile
ln -s afile merged/alink
stat --format "mod: %y ;acc: %x" merged/alink
echo "Trying to set date to 2020-01-02"
touch -h -d "2020-01-02 12:13:14" merged/alink
stat --format "mod: %y ;acc: %x" merged/alink
$umount `pwd`/merged
rm -r lower upper work merged

```

Versions:
kernel: 5.4.33-2-lts (arch)
fuse-overlayfs 1.0.0-1
fuse3 3.9.1-1

giuseppe:
thanks for the report.

Could you please try https://github.com/containers/fuse-overlayfs/pull/205 and see if it fixes the error you've seen?

PR #205, opened 2020-04-25
main: fix set mtime for a symlink

giuseppe:
Closes: https://github.com/containers/fuse-overlayfs/issues/204

### DIFF
--- a/main.c
+++ b/main.c
@@ -5015,7 +5015,12 @@ char **
 get_new_args (int *argc, char **argv)
 {
   int i;
-  char **newargv = malloc (sizeof (char *) * (*argc + 2));
+  char **newargv;
+
+  newargv = malloc (sizeof (char *) * (*argc + 2));
+  if (newargv == NULL)
+    error (EXIT_FAILURE, 0, "error allocating memory");
+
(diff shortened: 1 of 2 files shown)


progandy:
Thanks, that seems to work fine.

main.c
```diff
@@ -3646,7 +3646,7 @@ ovl_setattr (fuse_req_t req, fuse_ino_t ino, struct stat *attr, int to_set, stru
       if (fd >= 0)
         ret = futimens (fd, times);
       else
-        ret = utimensat (AT_FDCWD, path, times, AT_SYMLINK_NOFOLLOW);
+        ret = utimensat (AT_FDCWD, path, times, 0);
```
rhatdan, line 3649: So this will set the Link files time stamp Not the link it was pointing to?

giuseppe: we are using it to update the `/proc/PID/FD` entry which is a symlink but handled specially by the kernel, so we want to modify the symlink itself in this case.

It is very easy to get this wrong, so I've added some tests to avoid regressions in future.

rhatdan:
LGTM

Issue #211, opened 2020-05-26
get_node_up: Assertion `node->layer == get_upper_layer (lo)' failed.


v1bri:
I'm using fuse-overlayfs to share some (potentially large) git repositories among multiple worker processes. When performing some large merges I run into the following assertion error...

```
fuse-overlayfs: main.c:2768: get_node_up: Assertion `node->layer == get_upper_layer (lo)' failed.
```

Output of `--version`

```
$ fuse-overlayfs --version
fuse-overlayfs: version 0.7.8
FUSE library version 3.9.0
using FUSE kernel interface version 7.31
fusermount3 version: 3.9.0
```

### Steps to reproduce

This uses a rather large repository that is publicly accessible. Hopefully this is a sufficient test case, otherwise please let me know if you would like some help narrowing it down.

```
$ git clone https://github.com/analogdevicesinc/linux.git -b 2019_R1
$ mkdir upper && mkdir work && mkdir target
$ fuse-overlayfs -o lowerdir=$PWD/linux -o upperdir=$PWD/upper -o workdir=$PWD/work $PWD/target
$ git -C $PWD/target merge origin/master

<snip/>
error: Unable to add drivers/nvmem/Kconfig to database
warning: inexact rename detection was skipped due to too many files.
warning: you may want to set your merge.renamelimit variable to at least 5731 and retry the command.
warning: unable to unlink '/mnt/ssd/target/.git/index.lock': Transport endpoint is not connected

```

giuseppe:
thanks for the report.

Could you please verify if the patch here: https://github.com/containers/fuse-overlayfs/pull/212 solves the issue you are having?

We have not seen before the issue because we usually use fuse-overlays inside of a user namespace, and it runs with CAP_DAC_OVERRIDE, but in this case you are running without additional capabilities so it fails to set the xattr for a read only file.

I'd expect more similar failures when fuse-overlayfs is not used in a user namespace.

v1bri:
That does indeed solve the issue. Thanks!

PR #212, opened 2020-05-26
main: correctly report copyup error

giuseppe:
the copyup function returns the error code set in ret.  Make sure ret
has the correct return code if set_fd_origin fails.

Closes: https://github.com/containers/fuse-overlayfs/issues/211


rhatdan:
LGTM

Issue #213, opened 2020-06-14
[Errno 2] No such file or directory after some unknown steps


alxchk:
I have strange issue with fuse-overlayfs, which is perfectly reproducible, but I can't build any minimal sample for that.

I have very fun setup, with python 2.7 in wine in podman :)
Issue looks like this. 

During container creation, I install python packages using pip. In case I start this container (with already installed packages) once again, some random files began to disappear.

```
Installing collected packages: six, cython, pyjnius
  Attempting uninstall: six
    Found existing installation: six 1.15.0
    Uninstalling six-1.15.0:
      Successfully uninstalled six-1.15.0
  Rolling back uninstall of six
  Moving to c:\python27\lib\site-packages\six-1.15.0-py2.7.egg-info
   from c:\python27\lib\site-packages\~ix-1.15.0-py2.7.egg-info
  Moving to c:\python27\lib\site-packages\six.py
   from c:\users\root\temp\pip-uninstall-ua5iy4\six.py
  Moving to c:\python27\lib\site-packages\six.pyc
   from c:\users\root\temp\pip-uninstall-ua5iy4\six.pyc
ERROR: Could not install packages due to an EnvironmentError: [Errno 2] No such file or directory: 'C:\\Python27\\Lib\\site-packages\\six-1.15.0.dist-info\\LICENSE'
```

giuseppe:
can you please provide more information?  What version of fuse-overlayfs are you using?

What steps do you perform before hitting the issue?

alxchk:
I'm using latest from git. I observed this behaviour since the first usage of fuse-overlayfs. I'm using it since 0.7.7. 

As I said, I can't make minimal example of this. I will try to make something small, but problem usually arise in quite big project

alxchk:
Okay. Looks like these actions reproduces the issue in some form. Please note, I'm not sure which actions exactly triggers this. So this is a fat example.. 

( podman with fuse-overlayfs )

```
mkdir test; cd test; cat >script.sh << __EOF__
#!/bin/sh

PACKAGES_BUILD="netifaces msgpack-python u-msgpack-python construct bcrypt watchdog dukpy impacket zeroconf==0.19.1"
PACKAGES_BUILD="\$PACKAGES_BUILD pycryptodomex pycryptodome cryptography pyOpenSSL paramiko"

PACKAGES="rsa pefile rsa netaddr win_inet_pton netaddr pypiwin32 poster win_inet_pton dnslib"
[...]

PR #214, opened 2020-06-17
mkdir: if the destination exists atomically swap them

giuseppe:
if the destination already exists as it could not be properly cleaned
up, attempt to atomically swap the two directories and free the old
one.

Closes: https://github.com/containers/fuse-overlayfs/issues/213


giuseppe:
I confirm it fixes the issue, I will cut a new release once this is merged

rhatdan:
LGTM

PR #217, opened 2020-06-20
main: fix lookup for overflow uid

giuseppe:
we were mistakenly using the overflow GID also for UIDs lookups.  Not a big issue as they usually have the same value.


giuseppe:
thanks for the quick review :-)

PR #230, opened 2020-08-24
utils.c: fix file_exists_at on musl.

### LINKED ISSUE #174: Alpine linux: "cannot read upper dir: Out of memory"
Any time fuse-overlayfs is executed under any user (root, non-root or user namespace) on Alpine linux host, it fails with 
```
fuse-overlayfs: cannot read upper dir: Out of memory
```
Enabling the debug output via `-d` or `-o debug` doesn't produce any additional output.

Uname (Alpine 3.11.3):
```
Linux alpine 5.4.12-1-virt #2-Alpine SMP Thu, 16 Jan 2020 12:53:54 UTC x86_64 Linux
```
Note that Alpine uses _musl libc_. Version of _libfuse3_ is 3.9.0.

**Steps to reproduce:**
```
apk add make autoconf automake fuse3 fuse3-dev gcc gettext git libc-dev linux-headers
git clone https://github.com/containers/fuse-overlayfs.git
cd fuse-overlayfs
./autogen.sh
./configure
make
make install
```
```
service fuse start
mkdir /mnt/upper /mnt/lower /mnt/work /mnt/merged
fuse-overlayfs -o lowerdir=/mnt/lower,upperdir=/mnt/upper,workdir=/mnt/work /mnt/merged
```

As a side note, I haven't been able to compile statically either, however I haven't invested much time into it as it seems that it's the reason why [Dockerfile.alpine](https://github.com/containers/fuse-overlayfs/blob/master/Dockerfile.alpine) is still empty.

mgoltzsche:
using `fstatat` instead of `faccessat` for musl/alpine compatibility.
also adds missing memory reallocation, an alpine test and completes the alpine Dockerfile.

Closes #174.

Also see [my comment](https://github.com/containers/fuse-overlayfs/issues/174#issuecomment-679363668).

Credits go to @yaamai for his [patch](https://github.com/yaamai/fuse-overlayfs/commit/3733cfcb6936df685e62c5917493223aab700f5c) that fixed the string allocation and showed me where the `EINVAL` (invalid argument) error happens.


utils.c
```diff
@@ -103,7 +103,8 @@ safe_openat (int dirfd, const char *pathname, int flags, mode_t mode)
 int
 file_exists_at (int dirfd, const char *pathname)
 {
-  return faccessat (dirfd, pathname, F_OK, AT_SYMLINK_NOFOLLOW|AT_EACCESS);
+  struct stat buf;
+  return fstatat(dirfd, pathname, &buf, AT_SYMLINK_NOFOLLOW);
```
giuseppe, line 107: stat is slower than accessat.  Could we use it as a fallback instead?

```
int ret = faccessat ....;
if (ret < 0 && errno == EINVAL)
  {
    struct stat buf;
    return fstatat (dirfd, pathname, &buf, AT_SYMLINK_NOFOLLOW);
  }
return ret;
```

rhatdan: Wasn't this just hangking on musl/alpine?

mgoltzsche: @giuseppe I've added your fallback code, completed the Dockerfile and added an alpine test.

mgoltzsche: @rhatdan do you mean it is slightly slower on alpine then?
Do you think a preprocessor instruction would be more appropriate to use one or the other variant?

giuseppe (approved):
LGTM

utils.c
giuseppe: the patch LGTM, let's wait for @rhatdan to chime in as he had a comment

giuseppe:
ok let's merge and if there is anything more to fix, we can have a follow-up patch

PR #237, opened 2020-10-04
main: prevent ENAMETOOLONG for whiteout files

giuseppe:
adding the .wh. prefix could cause the lookup to cross the f_namemax limit and fail the lookup with ENAMETOOLONG.  If the lookup fails with ENAMETOOLONG then the whiteout file doesn't exist.

Also, prevent files longer than the new limit can be created, otherwise it won't be possible to create whiteout files.

Closes: https://github.com/containers/fuse-overlayfs/issues/236

### DIFF
--- a/main.c
+++ b/main.c
@@ -145,6 +145,7 @@ open_by_handle_at (int mount_fd, struct file_handle *handle, int flags)
 #define PRIVILEGED_OPAQUE_XATTR "trusted.overlay.opaque"
 #define PRIVILEGED_ORIGIN_XATTR "trusted.overlay.origin"
 #define OPAQUE_WHITEOUT ".wh..wh..opq"
+#define WHITEOUT_MAX_LEN (sizeof (OPAQUE_WHITEOUT))
 
 #if !defined FICLONE && defined __linux__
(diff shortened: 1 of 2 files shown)


rhatdan:
LGTM

main.c
```diff
@@ -145,6 +145,7 @@ open_by_handle_at (int mount_fd, struct file_handle *handle, int flags)
 #define PRIVILEGED_OPAQUE_XATTR "trusted.overlay.opaque"
 #define PRIVILEGED_ORIGIN_XATTR "trusted.overlay.origin"
 #define OPAQUE_WHITEOUT ".wh..wh..opq"
+#define WHITEOUT_MAX_LEN (sizeof (OPAQUE_WHITEOUT))
```
avikivity: Note that eats an extra character, since sizeof also counts the trailing NUL.

giuseppe: yeah, fortunately I've found out the bug before the PR was merged and fixed it in the last version (https://github.com/containers/fuse-overlayfs/pull/237/files#diff-2045016cb90d1e65d71c2407a2570927R148).  Another issue is that the opaque whiteout is not used as a prefix, so it is enough to account for `.wh.`

Issue #238, opened 2020-10-04
fuse-overlayfs very memory hungry


bear0330:
Hi, 

I have about 9 containers run python app on a 4G ram machine. After couple weeks later, fuse-overlayfs process for every container will eat about 10~15% memory. It seems like slowly increased. 
So after few couple weeks (maybe 2 weeks) I need to reboot or restart all my containers to free memory.
Even my python apps doesn't use that lot of memory, I think there might some memory leak in fuse-overlayfs or it just very memory hungry process.

Thanks.

giuseppe:
when you see the memory usage so high, could you try running as root `# sync; echo 3 > /proc/sys/vm/drop_caches`?

It might be caused by the cache that is kept in memory when there is no memory pressure, it is not handled by fuse-overlayfs but the kernel decides when it must be freed.

bear0330:
Thank you, but actually OOM will work if I don't reboot my machine or restart container to make memory usage down. If it is just about cache, kernel will free it under memory pressure situation right? 
I will see many fuse-overlayfs process use more than 10% of memory after my containers has been killed by oom.

giuseppe:
I've opened a PR to solve a potential leak, and be more aggressive on releasing inodes.

kawhicurry:
I meet the same problem with `fuse-overlayfs version: 0.7.2`. The `buffer+cache memory` slowly increase and after they reached the top, `used memory` increase rapidly until it consume all the memory. I check the `top` and found `fuse-overlayfs` consume over 50% of the memory. I haven't test the latest version but it might work.

PR #239, opened 2020-10-06
main: fix a potential leak and be more aggressive on releasing inodes

giuseppe:
if the reference held by the directory is the last one, it would miss to clean up the inode.  Fix it by calling do_forget.

Closes: https://github.com/containers/fuse-overlayfs/issues/238

### DIFF
--- a/fuse-overlayfs.1
+++ b/fuse-overlayfs.1
@@ -55,7 +55,7 @@ reading/writing files to the system.
 
 .PP
 The fuse\-overlayfs dynamic mapping is an alternative and cheaper way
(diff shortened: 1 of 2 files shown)


giuseppe:
the node/inode handling needs more cleanups, but I am not sure how much effort it is worth given the effort for rootless overlay in the kernel

rhatdan:
I believe we should continue development on fuse-overlay to add advanced features like shifting of file system and potentially integration with other types of storeage like Warf from the other day.

rhatdan:
LGTM

Issue #241, opened 2020-10-08
Add `squash-to-root` mount option to support Podman `ignore_chown_errors` + `additionalimagestore`


plopresti:
See https://github.com/containers/storage/issues/735 for background.

Feature request: Add a `squash-to-root` mount option for use with Podman's `ignore_chown_errors` and `additionalimagestore`. This should have two effects:

1) Map all users/groups in the lower layers to zero on the way in to the overlay, leaving them alone (or using the provided maps?) on the way out.

2) Ignore various errors from chown/fchown/etc. which naturally occur when squashing everything to root.

If you like, I can work up a pull request once we agree on the name and semantics.

Thanks!

plopresti:
@giuseppe OK, it's a single fchown() here:

https://github.com/containers/fuse-overlayfs/blob/323b01f17478aad992dc69e9195d83bb490008d6/main.c#L2500-L2505

It happens twice in my brief run, first here:

```
#0  fchown () at ../sysdeps/unix/syscall-template.S:81
#1  0x0000000000407651 in create_directory (lo=lo@entry=0x7fff0682b840,	dirfd=4, name=0x738630 "run", 
    times=times@entry=0x7fff0682a0a0, parent=0x71c430, xattr_sfd=9, uid=65534, gid=0, mode=16877, 
    set_opaque=false, st_out=0x0) at main.c:2509
#2  0x0000000000407a43 in create_node_directory () at main.c:418
#3  0x0000000000407ee3 in copyup () at main.c:2689
#4  0x0000000000408142 in get_node_up (lo=0x7fff0682b840, node=0x7385b0) at main.c:2840
#5  0x0000000000409fc2 in ovl_do_open () at main.c:3292
#6  0x000000000040a469 in ovl_create (req=0x729b70, parent=7573776, name=0x7f0d78eec048 ".containerenv", 
    mode=33261,	fi=0x7fff0682b5a0) at main.c:3471
#7  0x00007f0d78bcaa7d in do_create (req=<optimized out>, nodeid=<optimized out>, inarg=<optimized out>)
    at ../lib/fuse_lowlevel.c:1275
#8  0x00007f0d78bcd4f6 in fuse_session_process_buf_int (se=se@entry=0x7295e0, buf=buf@entry=0x7fff0682b760, 
    ch=ch@entry=0x0) at ../lib/fuse_lowlevel.c:2601
[...]

PR #243, opened 2020-10-08
Add squash_to_root mount option

plopresti:
See issue https://github.com/containers/fuse-overlayfs/issues/241

### DIFF
--- a/main.c
+++ b/main.c
@@ -218,6 +218,8 @@ static const struct fuse_opt ovl_opts[] = {
    offsetof (struct ovl_data, plugins), 0},
   {"xattr_permissions=%d",
    offsetof (struct ovl_data, xattr_permissions), 0},
+  {"squash_to_root",
+   offsetof (struct ovl_data, squash_to_root), 1},
   FUSE_OPT_END
 };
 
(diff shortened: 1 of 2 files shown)


main.c
```diff
@@ -2688,7 +2697,10 @@ create_directory (struct ovl_data *lo, int dirfd, const char *name, const struct
   if (ret < 0)
     goto out;
 
-  if (uid != lo->uid || gid != lo->gid || get_upper_layer (lo)->has_stat_override || get_upper_layer (lo)->has_privileged_stat_override)
+  if (!lo->squash_to_root
```
giuseppe, line 2700: could we move this check inside do_fchown?

plopresti: Done.

Would similar handling make sense for do_chown and do_fchownat?

main.c
```diff
@@ -218,6 +218,8 @@ static const struct fuse_opt ovl_opts[] = {
    offsetof (struct ovl_data, plugins), 0},
   {"xattr_permissions=%d",
    offsetof (struct ovl_data, xattr_permissions), 0},
+  {"squash_to_root",
```
giuseppe, line 221: should this be `squash_to_root=%d` or would it work as a boolean if we don't specify the type?

giuseppe (commented):
one question, and could you please squash :-) the commits as well?

main.c
plopresti: Seems to work as a Boolean. I was trying to follow the docs (https://libfuse.github.io/doxygen/structfuse__opt.html).

plopresti:
> one question, and could you please squash :-) the commits as well?

Done. Also I think we do want the same handling for chown and fchownat, just for consistency. So I did that.

plopresti:
@giuseppe Let me know whether this looks good or if there is something else I need to do. Thanks!

giuseppe (approved):
thanks, it looks good to me

plopresti:
Thank you!

PR #244, opened 2020-10-08
man: removing the duplicate description of workdir option

sectorsize512:
Signed-off-by: Vasily Tarasov <vtarasov@us.ibm.com>


giuseppe:
thanks for the patch.  Could you please also drop it from crun.1.md?

rhatdan:
Merging, and I will open a PR for crun.1

PR #246, opened 2020-10-09
main: create inodes with 0755 with xattr_permissions

giuseppe:
make sure the inodes are created with mode 0755 when using
xattr_permissions, otherwise the original permission could prevent
running on NFS.  Requires Linux 5.9 for NFS with xattr support.


rhatdan:
LGTM

PR #248, opened 2020-10-19
fix errno

progrunner17:
It would be better if error caused by calloc or malloc set ENOMEM as errno.

### DIFF
--- a/main.c
+++ b/main.c
@@ -2289,7 +2289,7 @@ ovl_do_readdir (fuse_req_t req, fuse_ino_t ino, size_t size,
   buffer = malloc (size);
   if (buffer == NULL)
     {
-      fuse_reply_err (req, ENOMEM);
+      fuse_reply_err (req, errno);
       return;
     }
   p = buffer;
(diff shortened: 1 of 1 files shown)


saschagrunert (approved):
LGTM

main.c
```diff
@@ -4607,7 +4607,7 @@ ovl_readlink (fuse_req_t req, fuse_ino_t ino)
   buf = malloc (current_size);
   if (buf == NULL)
     {
-      fuse_reply_err (req, ENOENT);
+      fuse_reply_err (req, ENOMEM);
```
rhatdan, line 4610: Shouldn't this just be `errno`?

rhatdan:
@giuseppe PTAL

main.c
giuseppe: yes, let's fix it to be just `errno`

progrunner17:
I fixed all occurrence of following code to use  `errno` instead of `ENOMEM`.

```c
  buf = malloc (current_size);
  if (buf == NULL)
    {
      fuse_reply_err (req, ENOMEM);
```

giuseppe (approved):
thanks, LGTM

PR #252, opened 2020-10-21
main: honor option volatile

giuseppe:
upstream Linux added an option "volatile" for overlay mounts that has
the same meaning as fsync=0 already supported by fuse-overlayfs.


rhatdan:
LGTM

Issue #253, opened 2020-10-28
xattr_permissions=2 && --userns=keep-id results in files created with wrong owner


shenghaoyang:
#### Versions
```terminal
╰─❯ cat /etc/os-release 
NAME="Arch Linux"
PRETTY_NAME="Arch Linux"
ID=arch
BUILD_ID=rolling
ANSI_COLOR="38;2;23;147;209"
HOME_URL="https://www.archlinux.org/"
DOCUMENTATION_URL="https://wiki.archlinux.org/"
SUPPORT_URL="https://bbs.archlinux.org/"
BUG_REPORT_URL="https://bugs.archlinux.org/"
LOGO=archlinux
╭─ ~                                                                                                     at 01:12:29
╰─❯ fuse-overlayfs --version
fuse-overlayfs: version 1.1.0
FUSE library version 3.10.0
using FUSE kernel interface version 7.31
fusermount3 version: 3.10.0

╭─ ~                                                                                                     at 01:12:39
╰─❯ conmon --version
conmon version 2.0.21
commit: 35a2fa83022e56e18af7e6a865ba5d7165fa2a4a

╭─ ~                                                                                                     at 01:12:42
╰─❯ runc --version
runc version 1.0.0-rc92
commit: ff819c7e9184c13b7c2607fe6c30ae19403a7aff
spec: 1.0.2-dev

╭─ ~                                                                                                     at 01:12:47
╰─❯ podman --version
podman version 2.1.1
```
#### `storage.conf`
```toml
[storage]
driver = "overlay"
[storage.options]
additionalimagestores=["/usr/local/share/containers"]
[storage.options.overlay]
mount_program="/usr/bin/fuse-overlayfs"
mountopt="xattr_permissions=2,threaded=1"
```

### Problem descripion
I've setup a shared additional image store on `/usr/local/share/containers`.

`contrib/fix-mode.py` was run on the directories to set the xattrs and permissions required: 

```terminal
╭─ ∅ /usr/lo/sh/c/overlay/f/diff                                                                         at 01:21:46
╰─❯ getfattr -d *
# file: bin
user.fuseoverlayfs.override_stat="0:0:555"

# file: boot
user.fuseoverlayfs.override_stat="0:0:555"

<SNIP>
```

Everything goes well without `--userns=keep-id`:
```terminal
╰─❯ podman container run --net=host --ipc=host --rm -it fedora:latest
[root@Shenghao-BRIXS /]# touch afile
[...]

PR #255, opened 2020-11-02
main: always remap ids when specified

giuseppe:
if a mapping is specified, make sure it is always honored, also when
using xattr permissions.

Closes: https://github.com/containers/fuse-overlayfs/issues/253


saschagrunert (approved):
LGTM

rhatdan:
LGTM

PR #256, opened 2020-11-02
main: open hidden files

giuseppe:
if a file is delete and not accessible from the file system, but it is
still referenced, we must be able to open and use it.


saschagrunert (approved):
LGTM

rhatdan:
LGTM

PR #257, opened 2020-11-06
main: honor user.containers.override_stat

giuseppe:
also honor user.containers.override_stat to override containers stat override as it is set by containers/storage.


giuseppe:
this is needed for https://github.com/containers/storage/pull/744

rhatdan:
LGTM

PR #258, opened 2020-11-07
main: block setting the user.containers. xattr

giuseppe:
it is reserved for internal usage with containers/storage.


rhatdan:
LGTM

mtrmac:
Thanks!

Issue #265, opened 2020-12-07
inplace sed fails with permission error  


sgotand:
I encounter an error like this.
```bash
$ head -n2 /etc/os-release 
NAME="Ubuntu"
VERSION="20.04 LTS (Focal Fossa)"
$ fuse-overlayfs -V
fuse-overlayfs: version 1.3
FUSE library version 3.10.1
using FUSE kernel interface version 7.31
fusermount3 version: 3.10.1
$ mkdir mnt lo up work
$ echo hello > lo/hello.txt
$ fuse-overlayfs -o workdir=work,upperdir=up,lowerdir=lo ./mnt
$ sed -i 's/hello/HELLO/' ./mnt/hello.txt
sed: preserving permissions for ‘./mnt/sedYeWmNl’: Permission denied
$
```
With strace, I found out that sed failed at fchown.
```
99011 openat(AT_FDCWD, "/home/vagrant/sed_test/mnt/sedhPumiV", O_RDWR|O_CREAT|O_EXCL, 0600) = 4
99011 umask(002)                        = 0700
99011 fcntl(4, F_GETFL)                 = 0x8002 (flags O_RDWR|O_LARGEFILE)
99011 fstat(3, {st_mode=S_IFREG|0777, st_size=6, ...}) = 0
99011 read(3, "HELLO\n", 4096)          = 6
99011 fstat(4, {st_mode=S_IFREG|000, st_size=0, ...}) = 0
99011 read(3, "", 4096)                 = 0
99011 fchown(4, 1000, 1000)             = -1 EACCES (Permission denied)
99011 fchown(4, -1, 1000)               = -1 EACCES (Permission denied)
```
With further investigation, this failure seems from this line. 
https://github.com/containers/fuse-overlayfs/blob/master/main.c#L3878-L3884

Although I tried to fix this bug, but it was too hard for me, 
because I have no idea why permission error occurs with same user
nor why `fuse_file_info *fi` is set NULL and this line is executed.

giuseppe:
fuse-overlayfs is really meant to be used from a user namespace.  While it works without a user namespace, there can be some issues not encountered when running with capabilities, e.g. when accessing directories.

Could you show what is the permission mask for "mnt lo up work" ?

progrunner17:
Thank you for your explanation。.

Originally, their mask is 0002, but this error also occured even when I changed the mask to 0000.

giuseppe:
[...]

PR #266, opened 2020-12-12
avoid permission error on setting attribute of a file without permission

progrunner17:
In ovl_setattr, target file is opened when `fd` cannot be get from `fi->fh`,
and this may fail due to the mode of the file. 

For example, `chmod`ing a file with mode `000` fails because a file with mode `000` cannot be opened,
as explained in #265.

In order to avoid this failure, required permissions are added temporally to open the file 
and the original mode is restored after `fd` of the file is got.

### DIFF
--- a/main.c
+++ b/main.c
@@ -3859,6 +3859,7 @@ ovl_setattr (fuse_req_t req, fuse_ino_t ino, struct stat *attr, int to_set, stru
     {
       mode_t mode = node->ino->mode;
       int dirfd = node_dirfd (node);
+      bool need_restore_mode = false;
 
       if (mode == 0)
         {
@@ -3871,6 +3872,20 @@ ovl_setattr (fuse_req_t req, fuse_ino_t ino, struct stat *attr, int to_set, stru
               return;
             }
           node->ino->mode = mode = st.st_mode;
+        }
+      ret = faccessat(dirfd, node->path, R_OK|W_OK, AT_SYMLINK_NOFOLLOW);
+      if (ret < 0)
+        {
+          if (errno == EACCES)
+            {
+              need_restore_mode = true;
+              ret = fchmodat(dirfd, node->path, node->ino->mode | S_IRUSR | S_IWUSR, 0);
+            }
+          if (ret < 0)
+            {
+              fuse_reply_err (req, errno);
+              return;
+            }
         }
 
       switch (mode & S_IFMT)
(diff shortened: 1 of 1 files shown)


main.c
```diff
@@ -3910,6 +3925,15 @@ ovl_setattr (fuse_req_t req, fuse_ino_t ino, struct stat *attr, int to_set, stru
           strconcat3 (path, PATH_MAX, get_upper_layer (lo)->path, "/", node->path);
           break;
         }
+      if (need_restore_mode)
+        {
+          ret = fchmodat(dirfd, node->path, node->ino->mode, 0);
```
giuseppe, line 3930: it doesn't restore the mode to the original one, we cannot assume the original mode was `0`

main.c
```diff
@@ -3872,6 +3873,20 @@ ovl_setattr (fuse_req_t req, fuse_ino_t ino, struct stat *attr, int to_set, stru
...
+            {
+              need_restore_mode = true;
+              ret = fchmodat(dirfd, node->path, node->ino->mode | S_IRUSR | S_IWUSR, 0);
+            }
+          if (ret < 0)
+            {
+              fuse_reply_err (req, errno);
+              return;
+            }
+        }
```
giuseppe, line 3889: it is an extra cost that is not necessary in most cases, especially when running in a user namespace.

Could we just detect the failure from `safe_open` and work directly on  the path when that happens?  If the file could not be opened, could we attempt directly `fchmodat(dirfd, node->path, desired_mode)` and similarly for other functionalities?

giuseppe (commented):
thanks for the PR.

I've left some comments, I am worried this mechanism is not atomic.  Could we achieve the same result differently?

main.c
progrunner17: The mode to be set by `fchmodat` here is `node->ino->mode` rather than `0`.
c.f. 
```c
int fchmodat(int dirfd, const char *pathname, mode_t mode, int flags);
```

progrunner17:
Actually, I submitted the first draft of this PR to progress the discussion though I have same concerns as you.
Although I have little insight into filesystems thus it may take time to make a better PR,
but I'll do my best.

rhatdan:
What is going on with this PR.

rhatdan:
@progrunner17 @giuseppe What should we do with this PR?

giuseppe:
@progrunner17 are you still working on this PR?

progrunner17:
I'm sorry but I don't have enough time for this PR now.
Let me close this PR, and reopen it when I need to tackle the issue again.

PR #268, opened 2020-12-22
main: add squash_to_uid and squash_to_gid

giuseppe:
add a more generic variant of squash_to_root.

### DIFF
--- a/fuse-overlayfs.1
+++ b/fuse-overlayfs.1
@@ -43,6 +43,11 @@ A directory used internally by fuse\-overlays, must be on the same file
 system as the upper dir.
 
 .PP
+\fB\-o workdir=workdir\fP
+A directory used internally by fuse\-overlays, must be on the same file
+system as the upper dir.
+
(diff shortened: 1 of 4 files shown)


giuseppe:
@rhatdan PTAL

fuse-overlayfs.1
```diff
@@ -99,6 +104,18 @@ $ stat \-c %u:%g merged/a merged/b
 .PP
 Those are the same IDs visible from outside the user namespace.
 
+.PP
+\fB\-o squash\_to\_root\fP
```
rhatdan, line 108: Why have two different options for this?  Why not squash_to_uid=0?

giuseppe: we already have it in a released version, so it would be a breaking change.

I can make `squash_to_root` obsolete though and we can remove it in the future

rhatdan: SGTM

PR #272, opened 2021-01-25
main: check if whiteout device already exists

giuseppe:
on newer kernels unprivileged users can create whiteout devices.  If
the whiteout device creation failed with EEXIST, check whether the
existing file is already a whiteout.

Closes: https://github.com/containers/fuse-overlayfs/issues/271


rhatdan:
LGTM

Issue #273, opened 2021-01-30
"mv: replace 'blah', overriding mode 0007?" even though 'blah' does not exist


kwshi:
**Description**

> **EDIT/TLDR**: I have narrowed down the issue to a minimal, reproducible example; [Containerfiles demonstrating the bug can be found here](https://github.com/kwshi/buildah-bug-reports/tree/main/opam-help/minimal).  Skip to [this comment](https://github.com/containers/buildah/issues/2951#issuecomment-770247464) for specific steps and observations; the following stuff is what I originally wrote while trying to narrow down the problem, but it may not be super interesting to you.

Honestly, I don't know where to file the issue, since it's so obscure and specific.  But this bug doesn't show up when I use Docker, which leads me to believe it _may_ be a buildah-specific issue.  But I have otherwise no idea what's going on.  Here goes...

While trying to build a container with [opam](https://opam.ocaml.org) and some other OCaml packages installed, I received the following error message from opam while it was attempting to download packages:
```
+ /usr/sbin/mv "/home/opam/.opam/download-cache/sha256/e2/e2c4e8230f7c96236503fd75f22bdbc263639971bf104509e446855ded35ae1e.tmp.part" "/home/opam/.opam/download-cache/sha256/e2/e2c4e8230f7c96236503fd75f22bdbc263639971bf104509e446855ded35ae1e.tmp"
- /usr/sbin/mv: replace '/home/opam/.opam/download-cache/sha256/e2/e2c4e8230f7c96236503fd75f22bdbc263639971bf104509e446855ded35ae1e.tmp', overriding mode 0007 (------rwx)?
```
However, closer inspection reveals that the file `mv` thinks it's _replacing_, namely `/home/opam/****ae1e.tmp`, doesn't actually exist--so it doesn't make sense to me why it raises this warning at all.  My best guess is that it maybe has something to do with bugs in the filesystem implementation that is causing `mv` to think it needs to override a file when there actually isn't.

Building the same containerfile with Docker proceeds fine, without any errors, leading me to further believe this issue seems to lie somewhere in buildah/podman.

**Steps to reproduce the issue:**

1. I wrote a [Containerfile](https://github.com/kwshi/buildah-bug-reports/blob/main/opam-help/Containerfile) that reproduces the issue for me when attempting to build with `buildah bud .` or `podman build .` (both reproduce the issue for me).
   
[...]

PR #274, opened 2021-02-01
fuse-overlayfs: fix renameat2(RENAME_NOREPLACE)

giuseppe:
when device whiteouts are created (supported for unprivileged users in
newer Linux kernels) make sure the RENAME_NOREPLACE flag is dropped
when renaming the file on top of an existing whiteout.

Closes: https://github.com/containers/fuse-overlayfs/issues/273


rhatdan:
LGTM

Issue #279, opened 2021-03-04
Creating a symlink doesn't clear existing whiteout file


andrewsali:
The following shows that if a file is removed from the merged directory (by creating a whiteout), the whiteout is not cleared when subsequently a symlink is created in the place of the original file:

```
salia@LAPTOP-8286Q9P4:~/ofs_test/lowerdir$ ls
a.txt
salia@LAPTOP-8286Q9P4:~/ofs_test$ cd mergedir/
salia@LAPTOP-8286Q9P4:~/ofs_test/mergedir$ ls
a.txt
salia@LAPTOP-8286Q9P4:~/ofs_test/mergedir$ mv a.txt a2.txt
salia@LAPTOP-8286Q9P4:~/ofs_test/mergedir$ touch a3.txt
salia@LAPTOP-8286Q9P4:~/ofs_test/mergedir$ ln -s a3.txt a.txt
salia@LAPTOP-8286Q9P4:~/ofs_test/mergedir$ ls
a.txt  a2.txt  a3.txt
salia@LAPTOP-8286Q9P4:~/ofs_test/mergedir$ cd ..
salia@LAPTOP-8286Q9P4:~/ofs_test$ cd upperdir/
salia@LAPTOP-8286Q9P4:~/ofs_test/upperdir$ ls -laht
total 8.0K
drwxr-xr-x 2 salia salia 4.0K Mar  4 14:50 .
lrwxrwxrwx 1 salia salia    6 Mar  4 14:50 a.txt -> a3.txt
-rw-r--r-- 1 salia salia    0 Mar  4 14:50 a3.txt
-rwx------ 1 salia salia    0 Mar  4 14:50 .wh.a.txt
drwxr-xr-x 6 salia salia 4.0K Mar  4 14:48 ..
-rw-r--r-- 1 salia salia    0 Mar  4 14:48 a2.txt
```

You can see that .wh.a.txt is there, so next time it is remounted (see below), the file suddenly disappears :(

```
salia@LAPTOP-8286Q9P4:~/ofs_test/mergedir$ ls
a2.txt  a3.txt
```

andrewsali:
Version information: this is reproducible with overlayfs binary built from
latest master commit.

giuseppe:
thanks for the report and the reproducer.

I've opened a PR here: https://github.com/containers/fuse-overlayfs/pull/280

PR #280, opened 2021-03-05
main: mark directory not loaded on rename

giuseppe:
if a whiteout was created as part of a rename operation, mark the directory content as not loaded so to avoid some optimizations when a whiteout file must be deleted.

Closes: https://github.com/containers/fuse-overlayfs/issues/279


rhatdan:
LGTM

Issue #182, opened 2020-02-13
NFS and subdirectories with many files


dhzavann:
I've been testing a bit by using an NFS mount as the lower layer, and local (initially empty) upper and workdirs. This seems to be a bit slow in some cases.

As one example, we have a directory which contains only around 15 entries; all directories. Each of these directories in turn have around 12 subdirectories (and so on). The initial listing of the parent directory containing only the 15 entries takes around 4 seconds in some informal testing.

The output from tcpdump seems to suggest that fuse-overlayfs is also recursing into the subdirectories. Querying the NFS mount directly does not result in the same type of recursion and is reasonably fast.

I'm guessing there's a good reason for doing this, but is there some flag or anything else that I'm missing or that could be added to skip this recursion?

giuseppe:
That looks like something fuse-overlayys should not do, but it was never really tested on network file systems.

Are you using fuse-overlayfs as unprivileged user?  Could you give more information on how you are invoking it?

dhzavann:
I'm using fuse-overlayfs as root in an unprivileged lxd container. I built fuse-overlayfs statically following the instructions a couple of hours ago.

I have tried mounting NFS both with nfusr inside the container, and with the kernel client bind mounted from the host through lxd. The issue manifests a bit differently between them and between runs (caching?), but I can trigger long waits in both cases.

An example session (slightly anonymized):

> \# NFS is mounted on /mnt/nfs
> time ls -lh /mnt/nfs/another/example/path | wc -l
> 10
>
> real	0m0.007s
> user	0m0.001s
> sys	0m0.006s
> 
> rm -rf /tmp/nfs
> mkdir -p /tmp/nfs/{upper,workdir,mount}
>
> sudo ./fuse-overlayfs -o uidmapping=33:1000:1,gidmapping=33:1000:1,lowerdir=/mnt/nfs,upperdir=/tmp/nfs/upper,workdir=/tmp/nfs/workdir /tmp/nfs/mount
> 
> time ls -lh /tmp/nfs/mount/another/example/path | wc -l
> 10
>
> real	0m27.898s
> user	0m0.006s
> sys	0m0.000s


The another/example/path contains only directories named after years (2012, 2013, ..). Below those are directories named after months (1-12). I can see in tcpdump expected queries, like: 

> NFS request xid 245309214 128 lookup fh Unknown/0100074D0001000000000000FF1E9CE6D01EA81E00000000000000000CBBF500 "2013"

[...]

PR #282, opened 2021-03-16
main: Add option static_nlink to set st_nlink to 1 for all directories

dhedberg:
This is a continuation of the discussion in #193 and solves #182.

We've been using (something like) this internally for almost a year, and for our limited use case it works fine.

Anything that relies on accurate values for st_nlink will obviously break if this option is enabled.


giuseppe (approved):
LGTM

Issue #289, opened 2021-04-16
File descriptor leak in ovl_rename_direct


LukasRupprecht:
I recently encountered a file descriptor leak in fuse-overlayfs. My current setup is a bit complicated and I haven't had a chance yet to create a minimal reproducible case but essentially I ran into this problem when running a fuse-overlayfs on top of another fuse-overlayfs with another file system in between the two.

I was able to pin it down to [this code path](https://github.com/containers/fuse-overlayfs/blob/6457fd46c0985d9789f9bd08670258af5142ae17/main.c#L4510) in `ovl_rename_direct`, which returns before closing `srcfd` and `destfd`. Adding a
```
close (srcfd);
close (destfd);
```
before returning fixed the problem for me.

giuseppe:
thanks for the report!

Could you verify if the following patch also fixes the issue you've seen?

```diff
diff --git a/main.c b/main.c
index 14ac260..97cb0ee 100644
--- a/main.c
+++ b/main.c
@@ -4400,8 +4400,8 @@ ovl_rename_direct (fuse_req_t req, fuse_ino_t parent, const char *name,
   struct ovl_data *lo = ovl_data (req);
   int ret;
   int saved_errno;
-  int srcfd = -1;
-  int destfd = -1;
+  cleanup_close int srcfd = -1;
+  cleanup_close int destfd = -1;
   struct ovl_node key;
   bool destnode_is_whiteout = false;
 
@@ -4586,20 +4586,12 @@ ovl_rename_direct (fuse_req_t req, fuse_ino_t parent, const char *name,
   node->loaded = 0;
 
   ret = 0;
-  goto cleanup;
+  fuse_reply_err (req, 0);
+  return;
 
  error:
   ret = -1;
-
- cleanup:
-  saved_errno = errno;
-  if (srcfd >= 0)
-    close (srcfd);
-  if (destfd >= 0)
-    close (destfd);
-  errno = saved_errno;
-
-  fuse_reply_err (req, ret == 0 ? 0 : errno);
+  fuse_reply_err (req, errno);
 }
 
 static void

```

LukasRupprecht:
Thanks for the quick response! I just tried the patch and it fixed the problem :-).

giuseppe:
thanks for confirming it, I've opened a PR: https://github.com/containers/fuse-overlayfs/pull/290

PR #290, opened 2021-04-16
main, rename: fix leak on error

giuseppe:
fix a fd leak on the error exit path.

Closes: https://github.com/containers/fuse-overlayfs/issues/289


rhatdan:
LGTM

Issue #287, opened 2021-04-13
Stale directory information after file creation


gicmo:
It seems newly created files are not visible for an already opened directory, a file system access pattern like following should result in the final `getdents64 ` having an entry for the newly created file, while currently it does not:

```
open(<PATH>, O_RDONLY|O_LARGEFILE|O_CLOEXEC|O_DIRECTORY) = 4.              # open the dir (fd 4)

openat(4, "a", O_WRONLY|O_CREAT|O_TRUNC|O_LARGEFILE|O_CLOEXEC, 0666) = 5.  # create file in dir (fd 5)
write(5, "a", 1)                        = 1
close(5) 

lseek(4, 0, SEEK_SET)                   = 0                                # rewind the directory (4)
getdents64(4, 0x7fda56ea0ad8 /* 2 entries */, 2048) = 48                   # re-check the directory
```

This was first reported as a bug against the bolt test suit [issue #171](https://gitlab.freedesktop.org/bolt/bolt/-/issues/171). I have isolated the code into a separate repo: [gicmo/fs-test](https://github.com/gicmo/fs-test), which contains the simple reproducer [`empty-dir.c`](https://github.com/gicmo/fs-test/blob/main/empty-dir.c) and the corresponding container recipes for Alpine and Fedora.

giuseppe:
thanks for the detailed report.

Opened a PR here: https://github.com/containers/fuse-overlayfs/pull/291

PR #291, opened 2021-04-16
main: move directory load at readdir time

giuseppe:
move the logic for loading the directory content at readdir time when
the offset is set to 0, so that any new file to the directory is
propagated.

Closes: https://github.com/containers/fuse-overlayfs/issues/287


rhatdan:
LGTM

PR #300, opened 2021-06-21
main: fix invalid access when filtering xattrs

giuseppe:
Fix an invalid access when filtering internal xattrs.
    
The size passed to memmove was longer than the remaining bytes to process.

As part of the fix: move the filtering logic to a separate function.

Closes: https://github.com/containers/fuse-overlayfs/issues/293


rhatdan:
LGTM

PR #313, opened 2021-07-28
fuse-overlayfs: fix read xattrs for devices

giuseppe:
always use llistxattr and lgetxattr for listing and reading xattrs so
that the open/openat2 call doesn't fail when accessing a device.

Closes: https://github.com/containers/fuse-overlayfs/issues/312


rhatdan:
LGTM

Issue #311, opened 2021-07-22
Get 'Invalid argument' Error when ls merged directory


JinVei:
Hi, I have a trouble in using fuse-overlayfs

 I try to use fuse-overlayfs in k8s container, But get 'Invalid argument' error when ls 'merged' directory that mounted by fuse-overlayfs.

What is wrong about it, And how to fix the problem?

Many thanks
```
root@pc-normal-003-bubimoktdos:/home/overlay# ls
ls: cannot access 'merged': Invalid argument
lowerdir  merged  up  workdir
root@pc-normal-003-bubimoktdos:/home/overlay# ls merged
ls: cannot access 'merged': Invalid argument
root@pc-normal-003-bubimoktdos:/home/overlay# ll
ls: cannot access 'merged': Invalid argument
total 0
drwxr-xr-x 6 root root  61 Jul 22 17:09 ./
drwxr-xr-x 1 root root 111 Jul 22 16:55 ../
drwxr-xr-x 2 root root  20 Jul 22 16:57 lowerdir/
d????????? ? ?    ?      ?            ? merged/
drwxr-xr-x 2 root root   6 Jul 22 16:55 up/
drwxr-xr-x 3 root root  18 Jul 22 16:56 workdir/
```
```
root@pc-normal-003-bubimoktdos:/home/overlay# fuse-overlayfs-x86_64 -o  lowerdir=lowerdir,upperdir=up,workdir=workdir merged -d
uid=unchanged
gid=unchanged
upperdir=/home/overlay/up
workdir=workdir
lowerdir=lowerdir
mountpoint=merged
plugins=<none>
fsync=enabled
FUSE library version: 3.4.1
unique: 1, opcode: INIT (26), nodeid: 0, insize: 56, pid: 0
INIT: 7.23
flags=0x0003fffb
max_readahead=0x00020000
   INIT: 7.27
   flags=0x0001f069
   max_readahead=0x00020000
   max_write=0x00020000
   max_background=0
   congestion_threshold=0
   time_gran=1
   unique: 1, success, outsize: 80
unique: 2, opcode: GETATTR (3), nodeid: 1, insize: 56, pid: 4493
ovl_getattr(ino=1)
   unique: 2, error: -22 (Invalid argument), outsize: 16
unique: 3, opcode: GETATTR (3), nodeid: 1, insize: 56, pid: 4494
ovl_getattr(ino=1)
   unique: 3, error: -22 (Invalid argument), outsize: 16
[...]

PR #316, opened 2021-08-04
direct: treat statx EINVAL as ENOSYS

giuseppe:
glibc has a fallback code on ENOSYS that can cause EINVAL.

Attempt the fallback on both errors.

Closes: https://github.com/containers/fuse-overlayfs/issues/311


rhatdan:
LGTM

PR #319, opened 2021-08-10
NEWS: tag 1.7.1

giuseppe:
- set FUSE_CAP_POSIX_ACL only when it is supported by FUSE.
- treat statx failure with EINVAL as ENOSYS, so that the fallback is attempted.


saschagrunert (approved):
LGTM

Issue #337, opened 2022-01-28
podman commit ⇒ error copying layers and metadata, cannot allocate memory`


Hi-Angel:
**Is this a BUG REPORT or FEATURE REQUEST?**

/kind bug

**Description**

I can no longer save a podman container I use for work. When I execute a `podman commit trusting_vaughan bds-ubuntu:18.04`, I get:

```
Error: error copying layers and metadata for container "df73b540564255d08e6f7523b5294d665ceac94664d70a0e806d09b68315bfa3": initializing source containers-storage:trusting_vaughan: error extracting layer "d23b8508e20c6ed50f1b2033698748aa2d8c59154cac882a4d63607c2cf0adfd": lstat /home/constantine/.local/share/containers/storage/overlay/4cc82349e6dd9aa8a5bf342d383c1ccfac19416507f80afcc66f55e4bd77e390/merged/root/.web_modules/v4/node_modules/node_modules: cannot allocate memory
```

I asked on IRC but haven't gotten much except that it is somehow related to RAM and that it might be a bug. So I'm reporting it here.

Amount of RAM I have: 8G RAM + 7.5G SWAP, so more than enough.

**Steps to reproduce the issue:**

Unfortunately steps require the container I have here.

1. Start the container with `podman run --rm -it bds-ubuntu:18.04 /bin/zsh`
2. Try to save it with `podman commit $(podman ps | perl -lane 'print @F[-1] if /bds-ubuntu:18.04/') bds-ubuntu:18.04`

**Describe the results you received:**

```
Error: error copying layers and metadata for container "df73b540564255d08e6f7523b5294d665ceac94664d70a0e806d09b68315bfa3": initializing source containers-storage:trusting_vaughan: error extracting layer "d23b8508e20c6ed50f1b2033698748aa2d8c59154cac882a4d63607c2cf0adfd": lstat /home/constantine/.local/share/containers/storage/overlay/4cc82349e6dd9aa8a5bf342d383c1ccfac19416507f80afcc66f55e4bd77e390/merged/root/.web_modules/v4/node_modules/node_modules: cannot allocate memory
```

**Describe the results you expected:**

No errors.

**Output of `podman version`:**

```
Version:      3.4.4
API Version:  3.4.4
Go Version:   go1.17.4
Git Commit:   f6526ada1025c2e3f88745ba83b8b461ca659933
Built:        Thu Dec  9 21:30:40 2021
OS/Arch:      linux/amd64
```

[...]

PR #338, opened 2022-01-31
main: fix lookup if underlying file is a symlink

giuseppe:
fix lookup if the underlying file is a symlink, while it is a directory on the upper layer.

Closes: https://github.com/containers/fuse-overlayfs/issues/337


rhatdan:
LGTM

Issue #348, opened 2022-05-03
sed inplace edit removes all permissions on NFS filesystem


Tuupertunut:
Related to #265 
Possibly related to #327

**OS:** Ubuntu 20.04, kernel 5.13.0-40-generic
**fuse-overlayfs version:** 1.8.2
**Fuse library version:** 3.4.1
**NFS version:** 4.1
**sed version:** 4.7

**How to reproduce:**

Run this code on an NFS file system

```bash
unshare --mount --user --map-root-user bash -c '
    mkdir lower upper work mount
    fuse-overlayfs -o squash_to_root -o lowerdir=lower,upperdir=upper,workdir=work mount
    echo abcde > mount/testfile
    sed -i s/a/b/ mount/testfile
    fusermount -uz mount
'
```

**Actual behavior:**

It will fail with a message like
`sed: preserving permissions for ‘mount/sedZUa3s8’: Permission denied`

The contents of the directory `upper` will look like this
```
$ ls -al upper
total 8
drwx------ 2 myuser mygroup 4096 May  3 20:18 .
drwxr-xr-x 6 myuser mygroup 4096 May  3 20:18 ..
---------- 1 myuser mygroup    6 May  3 20:18 testfile
-rwx------ 1 myuser mygroup    0 May  3 20:18 .wh.sedZUa3s8
```

The file `testfile` has lost all of its permissions. There is also an unnecessary whiteout file left (issue #327). The content of `testfile` has changed from abcde to bbcde so the sed command has done its job.

**Expected behavior:**

Repeating the same code on an ext4 filesystem

It succeeds without errors.

The contents of the directory `upper` will look like this
```
$ ls -al upper
total 12
drwx------ 2 myuser mygroup 4096 May  3 20:39 .
drwx------ 6 myuser mygroup 4096 May  3 20:39 ..
c--------- 1 myuser mygroup 0, 0 May  3 20:39 sedEwVgW3
-rw------- 1 myuser mygroup    6 May  3 20:39 testfile
```

The file `testfile` has the expected permissions and the expected content.

Kawashima-Azumi:
Not seeing this permission denied error with fuse-overlayfs-1.8.2 with both low and upper dirs from nfs3. Mount options are
[...]

PR #350, opened 2022-05-27
Fix setting attributes on file without permissions

### LINKED ISSUE #349: Cannot chmod file with no permissions on NFS filesystem
Continues from #348 

**OS:** Ubuntu 20.04, kernel 5.13.0-44-generic
**fuse-overlayfs version:** latest commit from main dedd8410ed0dd52150c98a9d666b875388f5395e
**Fuse library version:** 3.9.0
**NFS version:** 4.1
**chmod version:** 8.30

**How to reproduce:**

Run this code on an NFS file system

```bash
unshare --mount --user --map-root-user bash -c '
    mkdir lower upper work mount
    fuse-overlayfs -o squash_to_root -o lowerdir=lower,upperdir=upper,workdir=work mount
    touch mount/testfile
    chmod a-rwx mount/testfile
    chmod u+rw mount/testfile
    fusermount -uz mount
'
```

**Actual behavior:**

The second chmod command fails with an error message
`chmod: changing permissions of 'mount/testfile': Permission denied`

The file `upper/testfile` has no permissions afterwards
`---------- 1 myuser mygroup    0 May 25 03:36 testfile`

**Expected behavior:**

Repeating the same code on an ext4 filesystem displays no error messages.

The file `upper/testfile` has the expected permissions
`-rw------- 1 myuser mygroup    0 May 25 04:31 testfile`

Also running the code without fuse-overlayfs on NFS works as expected

```bash
touch testfile
chmod a-rwx testfile
chmod u+rw testfile
```

Tuupertunut:
Fixes #349 
Also works around #348 on systems with older sed

Since there is already handling for the `default` case where we use a path instead of a file descriptor, we can just as well fall back to it if opening a file descriptor fails.

Note that I'm not familiar with the codebase and possible side effects that this might have.

### DIFF
--- a/main.c
+++ b/main.c
@@ -3984,10 +3984,7 @@ ovl_setattr (fuse_req_t req, fuse_ino_t ino, struct stat *attr, int to_set, stru
         case S_IFREG:
           cleaned_up_fd = fd = TEMP_FAILURE_RETRY (safe_openat (dirfd, node->path, O_NOFOLLOW|O_NONBLOCK|(to_set & FUSE_SET_ATTR_SIZE ? O_WRONLY : 0), 0));
           if (fd < 0)
-            {
-              fuse_reply_err (req, errno);
-              return;
-            }
+            strconcat3 (path, PATH_MAX, get_upper_layer (lo)->path, "/", node->path);
           break;
 
         case S_IFDIR:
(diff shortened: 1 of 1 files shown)


Tuupertunut:
There might be a similar problem also with `setxattr`. It also fails with permission denied on files with no permissions.

rhatdan:
@giuseppe PTAL

giuseppe (approved):
LGTM

PR #352, opened 2022-06-01
main: ignore EOVERFLOW when copying xattrs

giuseppe:
the kernel returns EOVERFLOW if the rootid cannot be mapped in the
current user namespace when reading the file capabilities
(security.capabilities xattr).


rhatdan:
LGTM

Issue #353, opened 2022-06-06
Wrong flag value for RENAME_NOREPLACE


Tuupertunut:
I think the value for the flag RENAME_NOREPLACE is wrong in the source.

https://github.com/containers/fuse-overlayfs/blob/878cb0ccad071d4adda506db3147b3bcefdff27e/main.c#L127-L130

In the linux kernel source it is `1 << 0`

https://github.com/torvalds/linux/blob/f2906aa863381afb0015a9eb7fefad885d4e5a56/include/uapi/linux/fs.h#L50-L52

This isn't usually a problem as the constant is already defined, but a bug anyway.

flouthoc:
Actually this indeed looks wrong to me as well and this could also conflict with `RENAME_WHITEOUT` as well if both `RENAME_WHITEOUT` and `RENAME_EXCHANGE` are undefined https://github.com/containers/fuse-overlayfs/blob/main/main.c#L133 @giuseppe WDYT ?

giuseppe:
yes, it seems to have the wrong value.

Would you mind opening a PR to fix it?

PR #355, opened 2022-06-07
main: set correct value of `RENAME_NOREPLACE` ifndef `RENAME_EXACHANGE`

flouthoc:
Set correct value of `RENAME_NOREPLACE` when `RENAME_EXCHANGE` is not
already defined i.e use `1 << 0` instead of `1 << 2` which seems
incorrect.

Reference: https://github.com/torvalds/linux/blob/f2906aa863381afb0015a9eb7fefad885d4e5a56/include/uapi/linux/fs.h#L50-L52

Closes: https://github.com/containers/fuse-overlayfs/issues/353

### DIFF
--- a/main.c
+++ b/main.c
@@ -124,9 +124,12 @@ open_by_handle_at (int mount_fd, struct file_handle *handle, int flags)
 }
 #endif
 
+#ifndef RENAME_NOREPLACE
+# define RENAME_NOREPLACE (1 << 0)
+#endif
+
 #ifndef RENAME_EXCHANGE
 # define RENAME_EXCHANGE (1 << 1)
-# define RENAME_NOREPLACE (1 << 2)
 #endif
 
 #ifndef RENAME_WHITEOUT


flouthoc:
@giuseppe PTAL

rhatdan:
Can  you sort the constants so 1<<0 comes first.

main.c
```diff
@@ -126,7 +126,7 @@ open_by_handle_at (int mount_fd, struct file_handle *handle, int flags)
 
 #ifndef RENAME_EXCHANGE
 # define RENAME_EXCHANGE (1 << 1)
-# define RENAME_NOREPLACE (1 << 2)
+# define RENAME_NOREPLACE (1 << 0)
```
rhatdan, line 129: Theoretically should RENAME_NOREPLACE be in its own block.  You could have a plaform that defines RENAME_EXCHANGE but not RENAME_NOREPLACE.

flouthoc: yes that is possible in theory hence moved to its own block.

rhatdan:
LGTM

giuseppe (approved):
LGTM

Issue #354, opened 2022-06-06
Missing whiteout when renaming file over existing whiteout


Tuupertunut:
**How to reproduce:**

```bash
unshare --mount --user --map-root-user bash -c '
    mkdir lower upper work mount
    touch lower/testfile
    touch lower/testfile2
    fuse-overlayfs -o squash_to_root -o lowerdir=lower,upperdir=upper,workdir=work mount
    rm mount/testfile2
    touch mount/testfile
    mv mount/testfile mount/testfile2
    fusermount -uz mount
'
```

**Expected behavior:**

After the mv there should be a whiteout shadowing `testfile` in `upper`.

**Actual behavior:**

The whiteout is missing.

**Why this happens:**

If there is already a whiteout in the destination path, and `rename` is called with flag `RENAME_NOREPLACE`, there is a special handling in the code for this case, but it does not create a whiteout to the source path like in all other cases.

See this code:

https://github.com/containers/fuse-overlayfs/blob/878cb0ccad071d4adda506db3147b3bcefdff27e/main.c#L4636

In this branch there is no code for creating a whiteout like in all other branches.

flouthoc:
@Tuupertunut Could you please try the patch in PR https://github.com/containers/fuse-overlayfs/pull/356 this is still in draft since this needs review.

PR #356, opened 2022-06-07
create `whiteout` when renaming file over existing `whiteout`

flouthoc:
Create `whiteout` for cases if there is already a whiteout in the destination path
and rename is called with flag `RENAME_NOREPLACE`.

Close: https://github.com/containers/fuse-overlayfs/issues/354

### DIFF
--- a/main.c
+++ b/main.c
@@ -4631,9 +4631,17 @@ ovl_rename_direct (fuse_req_t req, fuse_ino_t parent, const char *name,
       ret = direct_renameat2 (srcfd, name, destfd,
                               newname, flags|RENAME_WHITEOUT);
     }
-      /* If the destination is a whiteout, just overwrite it.  */
+  /* If the destination is a whiteout, just overwrite it.  */
   if (ret < 0 && errno == EEXIST)
-    ret = direct_renameat2 (srcfd, name, destfd, newname, flags & ~RENAME_NOREPLACE);
+    {
+      ret = direct_renameat2(srcfd, name, destfd, newname, flags & ~RENAME_NOREPLACE);
+      if (ret >= 0)
+        {
+          ret = create_whiteout(lo, pnode, name, false, false);
+          if (ret < 0)
+            goto error;
+        }
+    }
   if (ret < 0)
     {
       ret = direct_renameat2 (srcfd, name, destfd,


flouthoc:
This fixes the reported issue but i have to discuss the patch first and it would be nice if @Tuupertunut could try this as well.

main.c
```diff
@@ -4631,9 +4631,12 @@ ovl_rename_direct (fuse_req_t req, fuse_ino_t parent, const char *name,
       ret = direct_renameat2 (srcfd, name, destfd,
                               newname, flags|RENAME_WHITEOUT);
     }
-      /* If the destination is a whiteout, just overwrite it.  */
-  if (ret < 0 && errno == EEXIST)
-    ret = direct_renameat2 (srcfd, name, destfd, newname, flags & ~RENAME_NOREPLACE);
+  /* If the destination is a whiteout, just overwrite it.  */
+  if (ret < 0 && errno == EEXIST) {
+    ret = direct_renameat2(srcfd, name, destfd, newname, flags & ~RENAME_NOREPLACE);
+    if (!(ret < 0))
```
giuseppe, line 4637: this looks weird.

Could we have `ret >= 0`?

main.c
```diff
@@ -4631,9 +4631,12 @@ ovl_rename_direct (fuse_req_t req, fuse_ino_t parent, const char *name,
...
+      create_whiteout(lo, pnode, name, false, true);
```
giuseppe, line 4638: don't we need to check the return code?

And why do we need `force_create`?

main.c
flouthoc: fixed.

Tuupertunut:
From the code it looks like this fixes the issue. However I have already done a larger change for the rename code that also encompasses this fix and I will make a PR for it, you may want to check that instead.

flouthoc:
> From the code it looks like this fixes the issue. However I have already done a larger change for the rename code that also encompasses this fix and I will make a PR for it, you may want to check that instead.

@Tuupertunut Sure we can close this one if that PR already addresses this.

Tuupertunut:
Done: #357

flouthoc:
Closing this one in favor of : https://github.com/containers/fuse-overlayfs/pull/357

PR #357, opened 2022-06-07
Refactor and fix rename code

### LINKED ISSUE #327: Unnecessary whiteout file when renaming.
When a file is renamed, a whiteout file must be created in the upper directory if any lower directory contains the source. If the lower directories do _not_ contain the source then a whiteout file is not necessary, but it does not functionally harm anything to create one. In the Linux kernel OverlayFS implementation, if a file is renamed and the source is not in a lower directory, no whiteout file is created; however, the fuse-overlayfs implementation does create one.

This discrepancy doesn't appear to break anything functionally, but it does interfere with reproducibility when running commands which rename randomly-named files that only ever exist in the upper directory. For example, using `sed --in-place` to modify a file that is only in the upper directory can result in a randomly-named whiteout file being created in the upper directory because the way sed implements in-place editing is to create a new, randomly-named file and rename it into place atomically.

Here is a short bash script to demonstrate this discrepancy:
```bash
#!/bin/bash
set -euo pipefail

[[ ${EUID} -ne 0 ]] && echo "Must run as root!" && exit 1

# Show useful system information.
uname -r
fuse-overlayfs --version
echo

function run_test {
  # Work in a temporary directory.
  TEST_DIR="$(mktemp --directory)"
  cd "${TEST_DIR}"

  # Set up the test using the desired method.
  mkdir lower upper work merged
  if [[ "${1}" == "overlayfs" ]]; then
    mount -t overlay overlay -o lowerdir=lower,upperdir=upper,workdir=work merged
  else
    fuse-overlayfs -o lowerdir=lower,upperdir=upper,workdir=work merged
  fi

  # Perform the test.
  echo "test" > merged/test
  sed --in-place 's/test/TEST/' merged/test

  # Report the results.
  echo "METHOD=${1}"
  ls -Al upper
  echo

  # Clean up.
  umount merged
  rm --recursive --force "${TEST_DIR}"
}

run_test overlayfs
run_test fuse-overlayfs
```

The following is the output of that script on my system:
```
5.14.14-arch1-1
fuse-overlayfs: version 1.7.1
FUSE library version 3.10.5
using FUSE kernel interface version 7.31
fusermount3 version: 3.10.5

METHOD=overlayfs
total 4
-rw-r--r-- 1 root root 5 Oct 23 22:34 test

METHOD=fuse-overlayfs
total 4
c--------- 1 root root 0, 0 Oct 23 22:34 sedKDg08c
-rw-r--r-- 1 root root    5 Oct 23 22:34 test
```

Here fuse-overlayfs has a `sedKDg08c` whiteout file, but Linux OverlayFS has no whiteout file. Note, unnecessary whiteout files are _not_ created when _removing_ files that only ever existed in the upper directory, this appears to apply only to renaming.

### LINKED ISSUE #354: Missing whiteout when renaming file over existing whiteout

Tuupertunut:
- Create source whiteout only when needed (fixes #327)
- Fix missing source whiteout when destination is whiteout (fixes #354)
- Try the EXCHANGE trick also when NOREPLACE is set
- Clean up the renaming code

Notes:
- I copied the `src_needs_whiteout` check from the `hide_node` function. In that function, the check is much longer but I can't come up with any reason why it has those extra parts.
- I don't know what the `pnode->loaded = 0` line does. It was there before and I didn't remove it.

### DIFF
--- a/main.c
+++ b/main.c
@@ -4538,11 +4538,15 @@ ovl_rename_direct (fuse_req_t req, fuse_ino_t parent, const char *name,
   if (node == NULL)
     goto error;
 
+  /* If NOREPLACE flag is given, check if we should throw an error now.
+     If not, just remove the flag as it might cause problems with replacing whiteouts later.  */
   if (flags & RENAME_NOREPLACE && destnode && !destnode->whiteout)
     {
       errno = EEXIST;
       goto error;
     }
+  else
+    flags = flags & ~RENAME_NOREPLACE;
 
   if (destnode)
     {
@@ -4597,17 +4601,6 @@ ovl_rename_direct (fuse_req_t req, fuse_ino_t parent, const char *name,
         }
     }
 
-  /* If the destnode is a whiteout, first attempt to EXCHANGE the source and the destination,
-   so that with one operation we get both the rename and the whiteout created.  */
-  if (destnode_is_whiteout)
-    {
-      ret = direct_renameat2 (srcfd, name, destfd, newname, flags|RENAME_EXCHANGE);
-      if (ret == 0)
-        goto done;
-
-      /* If it fails for any reason, fallback to the more articulated method.  */
-    }
-
   /* If the node is a directory we must ensure there is no whiteout at the
      destination, otherwise the renameat2 will fail.  Create a .wh.$NAME style
      whiteout file until the renameat2 is completed.  */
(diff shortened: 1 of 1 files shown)


main.c
```diff
@@ -4619,39 +4612,51 @@ ovl_rename_direct (fuse_req_t req, fuse_ino_t parent, const char *name,
...
   if (! can_mknod)
     {
       ret = -1;
       errno = EPERM;
     }
   else
-    {
-      ret = direct_renameat2 (srcfd, name, destfd,
-                              newname, flags|RENAME_WHITEOUT);
+            ret = direct_renameat2 (srcfd, name, destfd, newname, flags|RENAME_WHITEOUT);
```
giuseppe, line 4632: indentation seems off

giuseppe (commented):
could you please re-indent your code?  Otherwise LGTM and we can merge

main.c
Tuupertunut: That's VS Code or Git doing something weird. It was right when I committed it. 🙄

giuseppe (approved):
LGTM

Issue #364, opened 2022-07-27
Possible regression when lower layer is under the target mountpoint


flx42:
See https://github.com/NVIDIA/enroot/issues/130 for the full details and how it creates an almost-unkillable `fuse-overlafys`  process when the mountpoint is above the lower dir, Red Hat support mentioned that this is an invalid use case, but I went back and verified it worked fine in fuse-overlayfs v0.7 -> v1.6, and it started hanging since this seemingly unrelated commit in v1.7 so I wanted to verify if this pattern should not be used indeed:
```
$ git bisect bad
4ad759b35a054b6010b348755921977a9dcf6e0e is the first bad commit
commit 4ad759b35a054b6010b348755921977a9dcf6e0e
Author: Giuseppe Scrivano <gscrivan@redhat.com>
Date:   Wed Jul 28 13:03:53 2021 +0200

    fuse-overlayfs: fix read xattrs for devices
    
    always use llistxattr and lgetxattr for listing and reading xattrs so
    that the open/openat2 call doesn't fail when accessing a device.
    
    Closes: https://github.com/containers/fuse-overlayfs/issues/312
    
    Signed-off-by: Giuseppe Scrivano <gscrivan@redhat.com>

 direct.c                 | 20 ++------------------
 tests/fedora-installs.sh |  3 +++
 2 files changed, 5 insertions(+), 18 deletions(-)
```

The repro code (tested on Ubuntu 5.17 and 5.15 kernels):
```console
bash -eux << 'EOF'
        rootfs="/run/user/$(id -u)/overlay"
        mkdir -p "${rootfs}"/{lower,upper,work}
        curl -SsL https://cdimage.ubuntu.com/ubuntu-base/releases/22.04/release/ubuntu-base-22.04-base-amd64.tar.gz | tar -C "${rootfs}/lower" -xz
        fuse-overlayfs --version
        fuse-overlayfs -f -o "lowerdir=${rootfs}/lower,upperdir=${rootfs}/upper,workdir=${rootfs}/work" "${rootfs}" &
        sleep 5s
        ls "${rootfs}"
EOF
```

Output with fuse-overlayfs 1.5, it works fine:
```console
+ fuse-overlayfs --version
fuse-overlayfs: version 1.5
FUSE library version 3.10.5
using FUSE kernel interface version 7.31
fusermount3 version: 3.10.5
+ sleep 5s
[...]

PR #366, opened 2022-07-29
direct: use /proc/self/fd to read xattrs

giuseppe:
instead of using the lgetxattr and llistxattr system calls on the
entire file path, use the /proc/self/fd/$FD/$RELATIVE_PATH path
instead so that the lookup is relative to the lower dir file
descriptor that is already open.

Closes: https://github.com/containers/fuse-overlayfs/issues/364

### DIFF
--- a/direct.c
+++ b/direct.c
@@ -44,8 +44,13 @@ static int
 direct_listxattr (struct ovl_layer *l, const char *path, char *buf, size_t size)
 {
   char full_path[PATH_MAX];
-
-  strconcat3 (full_path, PATH_MAX, l->path, "/", path);
+  int ret;
+  ret = snprintf (full_path, sizeof (full_path), "/proc/self/fd/%d/%s", l->fd, path);
+  if (ret >= sizeof (full_path))
+    {
+      errno = ENAMETOOLONG;
+      return -1;
+    }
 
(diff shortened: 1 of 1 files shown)


direct.c
```diff
@@ -44,8 +44,13 @@ static int
 direct_listxattr (struct ovl_layer *l, const char *path, char *buf, size_t size)
 {
   char full_path[PATH_MAX];
-
-  strconcat3 (full_path, PATH_MAX, l->path, "/", path);
+  int ret;
+  ret = snprintf (full_path, sizeof (full_path), "/proc/self/fd/%d/%s", l->fd, path);
+  if (ret == PATH_MAX)
```
flouthoc, line 49: I am not sure but should this be `if (ret >= PATH_MAX)` ?

flouthoc (approved):
LGTM. Otherthan one doubt above.

direct.c
rhatdan: Probably should be sizeof(full_path)

giuseppe: yes, thanks.  Fixed to `>= sizeof (full_path)`

flx42:
I confirm that this patch fixes the issue I'm seeing in #364, thanks!

rhatdan:
lgtm

Issue #397, opened 2023-05-30
Extraneous "/proc seems to be mounted as readonly, it can lead to unexpected failures"


DrDaveD:
In one case on EL8 I am consistently getting an extraneous warning message from fuse-overlayfs-1.11:
```
/proc seems to be mounted as readonly, it can lead to unexpected failures"
```
I checked the [code](https://github.com/containers/fuse-overlayfs/blob/main/main.c#L319) and it appears to be to be looking at a statvfs structure which is never initialized.  I'm not sure what it is supposed to be doing or what the check is for, but it is apparently not doing what it is supposed to do.

giuseppe:
thanks for the report, opened a PR: https://github.com/containers/fuse-overlayfs/pull/398

PR #398, opened 2023-06-07
main: fix check for writeable /proc

giuseppe:
fix the check for writeable /proc to use statfs instead of statvfs, and use the unitialized statvfs structure to check for the writeable bit.

Closes: https://github.com/containers/fuse-overlayfs/issues/397


giuseppe:
@flouthoc PTAL

flouthoc (approved):
LGTM

PR #403, opened 2023-09-13
CI: release: enable reproducible builds

AkihiroSuda:
The binaries are now bit-for-bit reproducible (<https://reproducible-builds.org/>):
```bash
docker buildx build \
   -o /tmp/fuse-overlayfs-builds \
   --build-arg SOURCE_DATE_EPOCH=$(git log -1 --pretty=%ct) \
   --platform=amd64,arm64,arm,s390x,ppc64le,riscv64 \
   -f Containerfile.cross .
```

This ensures that the binaries are not built from any unpublished source.

- Implementation is similar to <https://github.com/rootless-containers/slirp4netns/compare/v1.2.1...06b1097203487955714dcc1d7638c769d3a5eb89>
- `/etc/apt/sources.list` is modified to use <http://snapshot.ubuntu.com> using <https://github.com/reproducible-containers/repro-sources-list.sh>.
- As <http://snapshot.ubuntu.com> is slow, `/var/cache/apt` is cached on GHA using <https://github.com/reproducible-containers/buildkit-cache-dance>.
- For testing sake, `.github/workflows/release.yaml` is now executed for every push events. The release artifacts are published only on a push event with a `v*` tag.

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -1,98 +1,82 @@
 name: Release
 on:
   push:
-    tags:
-    - 'test-cross-*'
-    - 'v*'
+  pull_request:
 
 jobs:
   release:
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-22.04
     steps:
-    - uses: actions/checkout@v2
-    - uses: docker/setup-buildx-action@v1
+    - uses: actions/checkout@v4
+      with:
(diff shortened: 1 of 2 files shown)


rhatdan:
@giuseppe PTAL

giuseppe (approved):
LGTM

PR #410, opened 2023-11-01
main: propagate xattrs permissions with copyup

giuseppe:
when a directory is copied up, propagate its mode if using xattrs permissions.

Closes: https://github.com/containers/fuse-overlayfs/issues/407


flouthoc (approved):
LGTM

Issue #438, opened 2025-03-24
v1.14 reporting incorrect version


DrDaveD:
The [version in configure.ac](https://github.com/containers/fuse-overlayfs/blob/v1.14/configure.ac#L2) was not updated to 1.14 for the v1.14 release.

PR #439, opened 2025-03-26
Update AC_INIT to 1.15-dev

hswong3i:
The version string should update as "1.14" when we have stable release:

    -AC_INIT([fuse-overlayfs], [1.13-dev], [giuseppe@scrivano.org])
    +AC_INIT([fuse-overlayfs], [1.14], [giuseppe@scrivano.org])

But since we had already get
https://github.com/containers/fuse-overlayfs/releases/tag/v1.14 since 2024-06-28, here we bump the version string to "1.15-dev" for next stable release.

Fixes https://github.com/containers/fuse-overlayfs/issues/438


giuseppe (approved):
LGTM

Issue #440, opened 2025-04-25
Doesn't handle escaped "':" characters in lower layer directory names


nalind:
The kernel's implementation accepts ":" in the names of lower directories if they're escaped, but fuse-overlayfs doesn't.
```bash
$ unshare -Urm --map-auto
# mkdir /tmp/test
# mount -t tmpfs tmpfs /tmp/test
# cd /tmp/test
# mkdir mkdir lower:1 upper:1 work:1 merged:1
# mount -t overlay -o lowerdir=lower\\:1,upperdir=upper\\:1,workdir=work\\:1 none merged:1
# umount merged:1
# fuse-overlayfs -o lowerdir=lower\\:1,upperdir=upper\\:1,workdir=work\\:1 merged:1
# fuse-overlayfs -o lowerdir=`pwd`/lower\\:1,upperdir=`pwd`/upper\\:1,workdir=`pwd`/work\\:1 `pwd`/merged:1
```
`mount` using the kernel succeeds, but `fuse-overlayfs` produces this error, give or take the current directory appearing in the incorrect path names:
```terminal
cannot resolve path lower
cannot load store (null) at lower
fuse-overlayfs: cannot read lower dirs: No such file or directory
```

giuseppe:
opened a PR:

- https://github.com/containers/fuse-overlayfs/pull/441

PR #441, opened 2025-04-28
main: Allow escaped colons in directory paths

giuseppe:
Allow directory paths specified for lowerdir, upperdir, workdir, and the mountpoint to contain colon characters.

Previously, colons were unconditionally treated as separators, making it impossible to use directories with colons in their names.

Closes: https://github.com/containers/fuse-overlayfs/issues/440

### DIFF
--- a/tests/fedora-installs.sh
+++ b/tests/fedora-installs.sh
@@ -2,17 +2,17 @@
 
 set -xeuo pipefail
 
-mkdir lower upper workdir merged
+mkdir lower:1 upper:2 workdir:3 merged
 
-fuse-overlayfs -o sync=0,lowerdir=lower,upperdir=upper,workdir=workdir,suid,dev merged
+fuse-overlayfs -o 'sync=0,lowerdir=lower\\:1,upperdir=upper\\:2,workdir=workdir\\:3,suid,dev' merged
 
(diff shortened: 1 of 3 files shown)


tests/fedora-installs.sh
```diff
@@ -2,18 +2,18 @@
 
 set -xeuo pipefail
 
-mkdir lower upper workdir merged
+mkdir lower:1 upper:2 workdir:3 merged:4
 
-fuse-overlayfs -o sync=0,lowerdir=lower,upperdir=upper,workdir=workdir,suid,dev merged
+fuse-overlayfs -o sync=0,lowerdir=lower\:1,upperdir=upper\:2,workdir=workdir\:3,suid,dev merged
```
nalind, line 7: Was this intended to use "merged:4"?

nalind:
From the description in #440, I don't think that the kernel needs escaping in mountpoint paths, since it hasn't been treating ":" in mountpoint paths as significant.

giuseppe:
> From the description in #440, I don't think that the kernel needs escaping in mountpoint paths, since it hasn't been treating ":" in mountpoint paths as significant.

thanks, fixed now

flouthoc (approved):
LGTM

Issue #442, opened 2025-04-29
gnulib changed: hash_delete -> hash_remove


bhaible:
Hi,

In 2021, Gnulib deprecated the `hash_delete` function. Please use `hash_remove` instead.

[replace-hash_delete.diff.gz](https://github.com/user-attachments/files/19966912/replace-hash_delete.diff.gz)

PR #443, opened 2025-04-29
main: use hash_remove instead of hash_delete

giuseppe:
the function is deprecated in gnulib

Closes: https://github.com/containers/fuse-overlayfs/issues/442

### DIFF
--- a/lib/hash.c
+++ b/lib/hash.c
@@ -1,6 +1,6 @@
 /* hash - hashing table processing.
 
-   Copyright (C) 1998-2004, 2006-2007, 2009-2023 Free Software Foundation, Inc.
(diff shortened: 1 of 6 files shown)


giuseppe:
@flouthoc PTAL

bhaible:
The title of this PR is wrong. It should be "use hash_remove instead of hash_delete". `hash_delete` is the old name, `hash_remove` the "new" name since 2021.

giuseppe:
> The title of this PR is wrong. It should be "use hash_remove instead of hash_delete". `hash_delete` is the old name, `hash_remove` the "new" name since 2021.

thanks for noticing, it was wrong in the commit message too.  Fixed now

flouthoc (approved):
LGTM

Issue #444, opened 2025-05-06
`fuse-overlayfs` returns `EEXISTS` on a `lookup` even when a `whiteout` file is found.


flouthoc:
`fuse-overlayfs` returns `EEXISTS` on a `lookup` even when a `whiteout` file is found.

Run buildah outer container
```
sudo ./podman run --privileged -it buildah bash
```
Inside the container create base image with following file
```
FROM alpine
RUN mkdir -p /base/test/test1
RUN touch /base/test/test1/test1-file
```
 
```
buildah build -t base -f Containerfile
```
 

Now build another container based of base and call this Containerfile2
```
FROM base
RUN mv /base/test/test1 /base/test/tmp
RUN cp -r /base/test/tmp /base/test/test1
```
```
buildah build -t base -f Containerfile2
```
 

You will see following error
```
[root@6aaf2ca101a8 mnt]# ./buildah build --authfile auth.json -t test -f Containerfile2
STEP 1/3: FROM base
STEP 2/3: RUN mv /base/test/test1 /base/test/tmp
STEP 3/3: RUN cp -r /base/test/tmp /base/test/test1
cp: can't open '/base/test/test1/test1-file': File exists
subprocess exited with status 1
subprocess exited with status 1
Error: building at STEP "RUN cp -r /base/test/tmp /base/test/test1": exit status 1
```

flouthoc:
I suspect this issue is happening because here the check does not honors if parent directory was marked as whiteout, irrespective of the state of parent directory it creates and return a new `ovl_node`

https://github.com/containers/fuse-overlayfs/blob/main/main.c#L2208

flouthoc:
To test I tried this patch

```diff
diff --git a/main.c b/main.c
index c06d0d1..5d40663 100644
--- a/main.c
+++ b/main.c
@@ -2204,8 +2204,8 @@ do_lookup_file (struct ovl_data *lo, fuse_ino_t parent, const char *name)
               wh_name = get_whiteout_name (name, &st);
               if (wh_name)
                 node = make_whiteout_node (path, wh_name);
-              else
-                node = make_ovl_node (lo, path, it, name, 0, 0, st.st_mode & S_IFDIR, pnode, lo->fast_ino_check);
+              //else
+              //  node = make_ovl_node (lo, path, it, name, 0, 0, st.st_mode & S_IFDIR, pnode, lo->fast_ino_check);
             }
           if (node == NULL)
             {
```

[...]

PR #445, opened 2025-05-07
main: lookup upperdir only for created directories

giuseppe:
always stop looking up at the upper layer for newly created directories so to not expose the lower directory content.

Closes: https://github.com/containers/fuse-overlayfs/issues/444


flouthoc (approved):
LGTM, this works

PR #457, opened 2026-03-25
Rewrite fuse-overlayfs in Rust

giuseppe:
Drop-in rewrite of fuse-overlayfs from C to Rust. Same CLI, same mount options, same on-disk format.


src/main.rs
```diff
@@ -0,0 +1,300 @@
...
+    // Initialize workdir
+    let workdir_fd = if let Some(ref wd) = config.workdir {
+        let fd = match sys::openat2::open_trusted(wd, libc::O_DIRECTORY, 0) {
+            Ok(fd) => fd,
+            Err(e) => {
+                eprintln!("fuse-overlayfs: cannot open workdir {}: {}", wd, e);
+                std::process::exit(1);
+            }
+        };
+        fd.as_raw_fd()
```
saschagrunert: Bug: `fd` is an `OwnedFd` that gets dropped at the end of this `if let` block, closing the OS file descriptor. `workdir_fd` then holds a dangling raw fd.

Keep the `OwnedFd` alive (e.g. store it alongside `workdir_fd`, or pass the `OwnedFd` directly to `OverlayFs::new`).

src/main.rs
```diff
@@ -0,0 +1,300 @@
...
+    let config = match config::parse_args(&args) {
+        Ok(c) => c,
+        Err(e) => {
+            eprintln!("fuse-overlayfs: {}", e);
+            std::process::exit(1);
+        }
+    };
+
+    // Read overflow IDs
+    let _overflow = mapping::OverflowIds::read();
```
saschagrunert: Dead code: `_overflow` is never used. `OverlayInner::new` creates its own `OverflowIds`. This can be removed.

src/main.rs
```diff
@@ -0,0 +1,300 @@
...
+            "ro" => fuse_options.push(fuser::MountOption::RO),
+            "suid" => fuse_options.push(fuser::MountOption::Suid),
+            "nosuid" => fuse_options.push(fuser::MountOption::NoSuid),
+            "dev" => fuse_options.push(fuser::MountOption::Dev),
+            "nodev" => fuse_options.push(fuser::MountOption::NoDev),
+            "exec" => fuse_options.push(fuser::MountOption::Exec),
+            "noexec" => fuse_options.push(fuser::MountOption::NoExec),
+            "atime" => fuse_options.push(fuser::MountOption::Atime),
+            "noatime" => fuse_options.push(fuser::MountOption::NoAtime),
+            "diratime" => fuse_options.push(fuser::MountOption::DirSync), // closest mapping
```
saschagrunert: Wrong mapping: `DirSync` != `diratime`. `diratime` controls directory access-time updates. This should be `CUSTOM("diratime".to_string())`.

src/sys/openat2.rs
```diff
@@ -0,0 +1,225 @@
...
+/// then `mknodat(fd_for_a/b, "c", ...)`.
+pub fn open_parent_safe<'a>(dirfd: RawFd, path: &'a str) -> FsResult<(Option<SafeFd>, &'a str)> {
+    match path.rfind('/') {
+        Some(pos) => {
+            let parent = &path[..pos];
+            let basename = &path[pos + 1..];
+            let parent_fd = safe_openat(
+                dirfd,
+                parent,
+                libc::O_DIRECTORY | libc::O_NOFOLLOW | libc::O_PATH,
```
saschagrunert: `O_NOFOLLOW` causes `ELOOP` if any component in the parent path is a symlink. Since `openat2` with `RESOLVE_IN_ROOT` already handles symlink containment, and the plain `openat` fallback path doesn't benefit from `O_NOFOLLOW` on intermediate components, drop `O_NOFOLLOW` and keep `O_DIRECTORY | O_PATH`.

src/sys/openat2.rs
```diff
@@ -0,0 +1,225 @@
...
+            OPENAT2_SUPPORTED.store(false, Ordering::Relaxed);
+            // fall through to fallback
+        } else if err == libc::EINVAL {
+            // fall through to fallback
+        } else {
+            return Err(FsError(err));
+        }
+    }
+
+    // Fallback to plain openat
```
saschagrunert: Silent fallback from `openat2` to plain `openat` drops the `RESOLVE_IN_ROOT` guarantee. Log a warning so users know the security constraint is absent.

Cargo.toml
```diff
@@ -0,0 +1,25 @@
...
+version = "2.0.0"
+edition = "2021"
+license = "GPL-2.0-or-later"
+description = "Overlay Filesystem in Userspace"
+
+[dependencies]
+fuser = { version = "0.17", features = ["abi-7-40"] }
+nix = { version = "0.29", features = ["fs", "dir", "mount", "user", "signal", "resource"] }
+libc = "0.2"
+libloading = "0.8"
```
saschagrunert: `libloading` is unused in the codebase. This can be removed.

src/sys/dir.rs
```diff
@@ -0,0 +1,147 @@
...
+/// # Safety invariant
+/// DIR* is not thread-safe, but this type is used exclusively within
+/// the OverlayFs big-lock mutex, so only one thread accesses it at a time.
+pub struct DirStream {
+    dir: *mut libc::DIR,
+}
+
+// Safety: DirStream is only used within the big-lock mutex in OverlayFs.
+// The mutex guarantees single-threaded access.
+unsafe impl Send for DirStream {}
```
saschagrunert, line 29: Comment says "never handed to another thread" but `Send` is exactly what enables cross-thread transfer. The real safety invariant is that `DIR*` is only used by one thread at a time (no concurrent `readdir` calls), which `&mut self` on `next()` already enforces.

src/mapping.rs
```diff
@@ -0,0 +1,205 @@
...
+        } else if id >= m.to && id < m.to + m.len {
+            return m.host + (id - m.to);
+        }
+    }
+
+    overflow_id
+}
+
+/// Convenience: map a UID from host to container.
+pub fn get_uid(
```
saschagrunert: `get_uid` and `get_gid` are never called. These can be removed.

src/node.rs
```diff
@@ -0,0 +1,866 @@
...
+
+impl InodeTable {
+    pub fn new() -> Self {
+        STAT_INODES.store(0, Ordering::Relaxed);
+        InodeTable {
+            table: FxHashMap::default(),
+            fuse_map: FxHashMap::default(),
+            same_device: true,
+            next_fallback: 0x8000_0000_0000_0000,
+        }
```
saschagrunert, line 320: With `ino_t_32` (32-bit inode numbers), `next_fallback` starts at `u32::MAX` and wraps around quickly, producing values 0 and 1 that collide with `FUSE_ROOT_ID`. Consider starting at 2 or skipping reserved values.

saschagrunert (commented):
Took a pass through the Rust rewrite. Left a few comments, mostly around dead code and a dangling fd in main.rs.

src/overlay.rs
```diff
@@ -0,0 +1,3944 @@
...
+            // Check if we already have a BackingId for this inode (concurrent open).
+            {
+                let backings = self.inode_backings.read();
+                if let Some((backing_id, _count)) = backings.get(&ino) {
+                    let fh = self.next_fh.fetch_add(1, Ordering::Relaxed);
+                    let backing_id = Arc::clone(backing_id);
+                    drop(backings);
+                    self.open_files.write().insert(fh, Arc::new(fd));
+                    self.fh_to_ino.write().insert(fh, ino);
+                    self.inode_backings.write().get_mut(&ino).unwrap().1 += 1;
```
saschagrunert: Race condition: the read lock on `inode_backings` is dropped on line 198, then a write lock is re-acquired here. Between the two, a concurrent `release()` can remove this entry, causing `.unwrap()` to panic.

Take a single write lock for the whole block instead of the read-then-write pattern.

src/copyup.rs
```diff
@@ -0,0 +1,418 @@
...
+        .ds
+        .openat(&path, libc::O_RDONLY | libc::O_NOFOLLOW, 0)?;
+
+    // Create temp file in workdir
+    let c_wd_name = crate::error::cstr(&wd_name)?;
+    let dfd = openat2::safe_openat(
+        workdir_fd,
+        &wd_name,
+        libc::O_CREAT | libc::O_WRONLY,
+        mode | 0o200,
```
saschagrunert, line 358: File is created with `mode | 0o200` to allow writing during copy, but `fchmod` is never called afterward to restore the original mode. A `0o444` source file ends up `0o644` after copy-up.

Add `fchmod(dfd, mode)` after `copy_data` completes.

src/sys/dir.rs
```diff
@@ -0,0 +1,149 @@
...
+            let entry = unsafe { libc::readdir(self.dir) };
+            if entry.is_null() {
+                return None;
+            }
+            // SAFETY: entry is non-null and valid for the lifetime of this readdir call.
+            let entry = unsafe { &*entry };
+            // SAFETY: d_name is a null-terminated C string in the dirent.
+            let name = unsafe {
+                std::ffi::CStr::from_ptr(entry.d_name.as_ptr())
+                    .to_string_lossy()
```
saschagrunert: `to_string_lossy()` replaces non-UTF-8 bytes with U+FFFD, making files with non-UTF-8 names (which are valid on Linux) inaccessible or silently mapped to the wrong name. This is a regression from the C version, which handles raw bytes.

Consider using `OsString` / `CString` throughout instead of `String`.

src/mapping.rs
```diff
@@ -0,0 +1,169 @@
...
+            return 0;
+        }
+    }
+
+    if mappings.is_empty() {
+        return id;
+    }
+
+    for m in mappings {
+        if direct {
```
saschagrunert, line 87: `m.host + m.len` can overflow `u32` when the mapping covers the top of the ID range. Use `id >= m.host && id - m.host < m.len` instead.

Same issue on line 90 with `m.to + m.len`.

src/overlay.rs
```diff
@@ -0,0 +1,3944 @@
...
+        }
+        // Apply to target
+        let _ = crate::sys::xattr::fsetxattr(target_fd, ACL_XATTR, &buf[..len], 0);
+    }
+
+    fn get_root(&self) -> &OvlNode {
+        // root_id is always valid; if not, we have a fatal invariant violation
+        self.nodes
+            .get(&self.root_id)
+            .expect("root node missing from arena")
```
saschagrunert: `expect()` will panic and crash the FUSE daemon if this invariant is ever violated. This goes against the project's own "no panic" rule in `AGENTS.md`. Consider returning a `Result` or using `fuse_err!`.

src/main.rs
```diff
@@ -0,0 +1,297 @@
+// fuse-overlayfs: Overlay Filesystem in Userspace
+//
+// Copyright (C) 2018 Giuseppe Scrivano <giuseppe@scrivano.org>
+// Copyright (C) 2018-2020 Red Hat Inc.
+//
+// SPDX-License-Identifier: GPL-2.0-or-later
+
+#![allow(dead_code)]
```
saschagrunert: `#![allow(dead_code)]` at crate level suppresses warnings for all unused code, hiding real dead code (e.g. the `get_uid`/`get_gid` functions flagged in the previous review). Consider removing this and cleaning up the actual dead code.

Cargo.toml
```diff
@@ -0,0 +1,24 @@
+[package]
+name = "fuse-overlayfs"
+version = "2.0.0"
+edition = "2021"
+license = "GPL-2.0-or-later"
+description = "Overlay Filesystem in Userspace"
+
+[dependencies]
+fuser = { version = "0.17", features = ["abi-7-40"] }
```
saschagrunert: Consider pinning dependencies to their exact patch versions (e.g. `fuser = "0.17.1"`, `nix = "0.29.1"`, etc.). For a filesystem, a semver-compatible but untested patch bump could introduce subtle behavioral changes or regressions. Pinning to known-good patch versions, combined with the `Cargo.lock`, gives more control over when updates are adopted.

saschagrunert:
This is a good chance to enhance the docs for this project, too.

saschagrunert (commented):
Second round of review. Previous comments were mostly addressed; these are the remaining and new findings.

Cargo.toml
```diff
@@ -0,0 +1,24 @@
+[package]
+name = "fuse-overlayfs"
+version = "2.0.0"
+edition = "2021"
```
saschagrunert: Consider updating to `edition = "2024"`. The 2024 edition enables `unsafe_op_in_unsafe_fn` by default, which would add an extra safety layer in the `src/sys/` modules where most of the unsafe code lives. It also enables other useful lints like `rust_2024_compatibility`.

src/sys/dir.rs
```diff
@@ -0,0 +1,149 @@
...
+                    .into_owned()
```
saschagrunert: This `to_string_lossy()` silently mangles non-UTF-8 filenames, which is a regression from the C version that handled raw bytes. This is a systemic issue: `RawDirEntry.name`, `OvlNode.name`, and the children map keys in `node.rs` all use `String`. They should use `OsString` (or `CString`/`Vec<u8>`) to preserve arbitrary POSIX filenames.

src/sys/openat2.rs
```diff
@@ -0,0 +1,194 @@
...
+// safe_openat: openat2 with RESOLVE_IN_ROOT to prevent path traversal.
+// SafeFd: newtype that can only be constructed through safe_openat.
+
+use crate::error::{FsError, FsResult};
+use std::os::fd::{AsRawFd, FromRawFd, OwnedFd, RawFd};
+
+const RESOLVE_IN_ROOT: u64 = 0x10;
+
+#[cfg(target_arch = "x86_64")]
+const SYS_OPENAT2: libc::c_long = 437;
```
saschagrunert: These per-arch `#[cfg]` blocks all define `SYS_OPENAT2 = 437` (same value on every architecture). A single `const SYS_OPENAT2: libc::c_long = 437;` would suffice, or better yet use `libc::SYS_openat2` which the `libc` crate already exports. The current approach silently fails to compile on unlisted architectures (e.g. `loongarch64`) with no error message.

src/config.rs
```diff
@@ -0,0 +1,538 @@
...
+    pub uid_str: Option<String>,
+    pub gid_str: Option<String>,
+    pub uid_mappings: Vec<IdMapping>,
+    pub gid_mappings: Vec<IdMapping>,
+
+    // Numeric options
+    pub timeout: f64,
+    pub threaded: i32,
+    pub fsync: i32,
+    pub fast_ino_check: i32,
```
saschagrunert: Several fields here use `i32` where more precise types would be appropriate:
- `fsync`, `writeback`, `disable_xattrs`, `fast_ino_check`, `threaded` are used as booleans (`!= 0`) but typed as `i32`.
- `squash_to_uid` / `squash_to_gid` use `-1` as a sentinel for "not set" instead of `Option<u32>`.

Using `bool` and `Option<u32>` respectively would make the intent clearer and prevent bugs from arithmetic on what are logically boolean/optional values.

.github/workflows/rust-test.yaml
```diff
@@ -0,0 +1,186 @@
...
+          ~/.cargo/registry
+          ~/.cargo/git
+          target
+        key: ${{ runner.os }}-cargo-${{ hashFiles('Cargo.lock') }}
+        restore-keys: |
+          ${{ runner.os }}-cargo-
+
+    - name: Build
+      run: cargo build --release
+
```
saschagrunert, line 36: There is no `cargo clippy` or `cargo fmt -- --check` step, despite `AGENTS.md` requiring zero warnings and formatted code. Adding these as CI steps would enforce those requirements.

.github/workflows/rust-test.yaml
```diff
@@ -0,0 +1,186 @@
...
+      uses: actions/checkout@v4
+
+    - name: Install Rust
+      uses: dtolnay/rust-toolchain@stable
+      with:
+        targets: ${{ matrix.target }}
+
+    - name: Install cross
+      run: cargo install cross --git https://github.com/cross-rs/cross
+
```
saschagrunert, line 139: `cargo install cross --git https://github.com/cross-rs/cross` installs from HEAD, which is not reproducible and could break without warning. Pin to a specific tag or revision, e.g. `cargo install cross --git https://github.com/cross-rs/cross --tag v0.2.5`.

giuseppe:
@saschagrunert thanks for the review!  I will ping you once I've addressed all the comments (I am pushing just to trigger the CI)

giuseppe:
comments addressed!

giuseppe:
@mheon FYI

giuseppe:
@saschagrunert are you fine to move forward with this one?

src/overlay.rs
```diff
@@ -0,0 +1,3935 @@
...
+        }
+
+        if let Some(new_size) = size {
+            let result = if fd >= 0 {
+                crate::sys::io::ftruncate(fd, new_size as i64)
+            } else {
+                match safe_parent(layer.ds.root_fd(), &path) {
+                    Ok((_guard, dirfd, c_name)) => {
+                        let full_path =
+                            format!("/proc/self/fd/{}/{}", dirfd, c_name.to_string_lossy());
```
saschagrunert: `to_string_lossy()` replaces non-UTF-8 bytes with U+FFFD, which would make this path point to the wrong file. Use `openat2::proc_fd_path()` (which already exists for this purpose) to build the path as raw bytes, then convert to `CString`.


```rust
                    Ok((_guard, dirfd, c_name)) => {
                        let full_path =
                            crate::sys::openat2::proc_fd_path(dirfd, c_name.to_bytes());
                        match CString::new(full_path) {
                            Ok(c) => crate::sys::fs::truncate(&c, new_size as i64),
                            Err(_) => Err(FsError(libc::EINVAL)),
                        }
                    }
```

saschagrunert (approved):
Just one issue, otherwise LGTM

src/overlay.rs
giuseppe: thanks!  Fixed it and pushed a new version

giuseppe:
@saschagrunert can I merge now? :-)

saschagrunert (approved):
/lgtm

giuseppe:
@saschagrunert @lsm5 any suggestions on how to make a release now?  Shall I just publish the git archive?

lsm5:
> @saschagrunert @lsm5 any suggestions on how to make a release now? Shall I just publish the git archive?

@giuseppe any particular concerns to account for? If not, I'd say just go with the default Github release creation process with the relevant git tag.

FWIW, I don't see any packit jobs in upstream, I did add a packit config to Fedora dist-git, but it's either not been tried yet or not working, as I don't see any updates in dist-git from Packit.

(Related: Btw, given this is a rewrite in another language, I don't know off-hand if we can continue using the existing fuse-overlayfs fedora package for updates. Might need to check with packaging committee on this. cc @mathstuf)

Luap99:
@giuseppe FYI, IIRC our debian packager wants a crate published on crates.io (at least for nv/av we had to do this)

In general it would be good to publish this sooner than later on crates.io to reserve the crate name for you, https://doc.rust-lang.org/cargo/reference/publishing.html#before-publishing-a-new-crate

mathstuf:
Not sure why I was pinged here; I don't maintain `fuse-overlayfs` in Fedora. But as a Fedora maintainer otherwise, my main concern would be with compatibility as far as everything else is concerned.

lsm5:
> Not sure why I was pinged here; I don't maintain `fuse-overlayfs` in Fedora. But as a Fedora maintainer otherwise, my main concern would be with compatibility as far as everything else is concerned.

@mathstuf I gave you commit access per https://src.fedoraproject.org/rpms/fuse-overlayfs/pull-request/6#comment-289565 . Anyway, if you prefer not to be on it, I'll not bother you here and drop you from there.

mathstuf:
Ah, I had forgotten about that :) . I mostly pay attention to Bugzilla for this. But either way, I shared my view on it.

giuseppe:
> But as a Fedora maintainer otherwise, my main concern would be with compatibility as far as everything else is concerned.

compatibility-wise, it should be a drop in replacement without any visible difference, anything that doesn't work will be treated as a bug.

mathstuf:
Sounds like something for Rawhide and, perhaps, a (minor?) Change Proposal for awareness.

Cargo.toml
```diff
@@ -0,0 +1,24 @@
+[package]
+name = "fuse-overlayfs"
+version = "2.0.0"
+edition = "2024"
+license = "GPL-2.0-or-later"
+description = "Overlay Filesystem in Userspace"
+
+[dependencies]
+fuser = { version = "0.17.0", features = ["abi-7-40"] }
+nix = { version = "0.29.0", features = ["fs", "dir", "mount", "user", "signal", "resource"] }
```
cgwalters, line 10: This is an old version of nix, if there's LLMs involved here, you have to be on guard about this as sometimes they'll synthesize this stuff from their training cutoff.

For sure we'll need renovate or dependabot setup.

Also though more importantly, IMO rustix is generally superior to nix and has already hit 1.0 and is already in the depchain of other things.

giuseppe: AFAICS nix is used by fuser.

Anyway, let's give it a try: https://github.com/containers/fuse-overlayfs/pull/463

cgwalters: Yeah it's not uncommon to have both in the depchain, and nix is OK, maybe at some point they'll rebase on top of rustix.

PR #458, opened 2026-04-06
feat: introduce Android bionic libc compatibility layer

subhu2008:
- Implements android_compat.h to map missing POSIX functions in bionic libc.
- Adds compiler detection to prevent breaking standard GNU libc desktop builds.
- Includes Termux build documentation in README.


giuseppe:
would https://github.com/containers/fuse-overlayfs/pull/457 solve the issue you're having?

subhu2008:
Nope, #457 doesn't solve it

PR #462, opened 2026-05-07
add FUSE_OVFS_IOC_REFRESH_DIR ioctl to refresh directory cache

wangjia184:
Summary                                                                                                                                                                                                                               
                                                                                                                                                                                                                                        
  Add a custom FUSE ioctl (FUSE_OVFS_IOC_REFRESH_DIR) that clears and re-scans a directory's cached children on demand, so that externally created .wh.* whiteout files and other changes to the backing directories become visible
  without remounting.                                                                                                                                                                                                                   
                                                            
  Motivation                                                                                                                                                                                                                            
                                                            
  When fuse-overlayfs scans a directory for the first time, it caches the child entries and marks the directory as "loaded." Subsequent reads skip the disk scan and return cached results. This means:                                 
                                                            
  - A .wh.* whiteout file created directly in the upper layer after the scan won't hide the corresponding file.                                                                                                                         
  - A file added directly to the lower layer won't appear.  
  - A file removed from the lower layer won't disappear.                                                                                                                                                                                
                                                                                                                                                                                                                                        
  The ioctl gives users a way to explicitly refresh a directory when external modifications have been made.

### DIFF
--- a/src/node.rs
+++ b/src/node.rs
@@ -263,6 +263,20 @@ impl OvlNode {
             whiteouts.insert(name);
         }
     }
+
+    /// Clear all children and whiteouts. Used by cache invalidation to
+    /// force a full re-scan of the directory.
+    pub fn clear_children(&mut self) {
+        if let DirState::Dir {
+            children,
+            whiteouts,
+            ..
+        } = &mut self.dir_state
+        {
+            children.clear();
+            whiteouts.clear();
+        }
+    }
 }
 
 impl Drop for OvlNode {
--- a/src/overlay.rs
+++ b/src/overlay.rs
@@ -100,6 +100,12 @@ pub struct OverlayFs {
     notifier: Arc<OnceLock<fuser::Notifier>>,
 }
 
+/// Custom ioctl command: clear and re-scan a directory's cached children.
+/// Encoded as _IO('f', 0x66) — direction=none, size=0, type='f'(0x66), nr=0x66.
+/// Immediately rebuilds the directory listing, picking up externally created
+/// whiteout files and other changes.
+const FUSE_OVFS_IOC_REFRESH_DIR: libc::Ioctl = (b'f' as libc::Ioctl) << 8 | 0x66;
+
 struct OverlayInner {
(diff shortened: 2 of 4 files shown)


giuseppe:
you need to sign off your commit, and please use a real name for that

src/node.rs
```diff
@@ -263,6 +263,20 @@ impl OvlNode {
             whiteouts.insert(name);
         }
     }
+
+    /// Clear all children and whiteouts. Used by cache invalidation to
+    /// force a full re-scan of the directory.
+    pub fn clear_children(&mut self) {
```
giuseppe, line 269: can't we just use `mark_unloaded`?

src/overlay.rs
```diff
@@ -3653,6 +3659,36 @@ impl Filesystem for OverlayFs {
         let is_set = cmd_ioctl == libc::FS_IOC_SETVERSION || cmd_ioctl == libc::FS_IOC_SETFLAGS;
         let is_get = cmd_ioctl == libc::FS_IOC_GETVERSION || cmd_ioctl == libc::FS_IOC_GETFLAGS;
 
+        // Custom command: clear and re-scan a directory's children.
+        if cmd_ioctl == FUSE_OVFS_IOC_REFRESH_DIR {
```
giuseppe, line 3663: we need to make sure the caller has appropriate permissions, can a process without privileges call the ioctl?

src/node.rs
wangjia184: In fact, I tried `mark_unloaded` but that does not work for `.wh.*` changes.  And I dont want to change the load part.

src/overlay.rs
wangjia184: There is no special permission required by `FUSE_OVFS_IOC_REFRESH_DIR`.

giuseppe: it can be abused by an unprivileged user (e.g. uid != 0 inside the container), to cause a lot of load on the file system as it invalidates the entire directory.

PR #463, opened 2026-05-07
fuse-overlayfs: use rustix instead of nix

giuseppe:


giuseppe:
@cgwalters does this look good to you?

cgwalters (approved):

src/main.rs
```diff
@@ -209,11 +209,18 @@ fn main() {
...
+    let rlim = getrlimit(Resource::Nofile);
+    if let Some(hard) = rlim.maximum {
+        if let Err(e) = setrlimit(
+            Resource::Nofile,
+            Rlimit {
+                current: Some(hard),
+                maximum: Some(hard),
+            },
+        ) {
+            eprintln!("fuse-overlayfs: cannot set nofile rlimit: {}", e);
```
cgwalters, line 222: Not a new thing but error logging via raw `eprintln!` i think is pretty crude; typically want tracing crate or so.

giuseppe: thanks for the suggestion, I'll address that separately

PR #464, opened 2026-05-20
fuse-overlayfs: replace eprintln! with log crate macros

giuseppe:
Move env_logger initialization after config parsing so --debug can set the log level.  Use error!/warn!/debug! instead of raw eprintln! throughout the codebase, with a custom format that preserves the "fuse-overlayfs: " prefix.

@cgwalters FYI

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -3,6 +3,7 @@
 // CLI option parsing. Replicates the exact -o key=value syntax from the C version.
 
 use crate::mapping::{self, IdMapping};
+use log::warn;
 
 /// All configuration parsed from the command line.
 #[derive(Debug)]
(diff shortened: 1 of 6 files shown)


giuseppe:
@saschagrunert PTAL

src/config.rs
```diff
@@ -178,7 +178,7 @@ pub fn parse_args(args: &[String]) -> Result<OverlayConfig, String> {
         }
 
         if arg.starts_with('-') {
-            eprintln!("unknown argument ignored: {}", arg);
+            log::warn!("unknown argument ignored: {}", arg);
```
saschagrunert: nit: for consistency I'd either use either `warn!(...)` or `log::warn!(...)` across the whole code base.

saschagrunert (approved):
Just a nit, otherwise LGTM

src/config.rs
giuseppe: thanks, fixed now

saschagrunert (approved):

PR #466, opened 2026-06-09
NEWS: tag 1.17

giuseppe:


giuseppe:
@mosskappa the original fix is broken, are you fine if I modify your patch?  Otherwise I'll replace it with a different one

mosskappa:
Yes, please feel free to modify my patch, or replace it if that is safer for the 1.17 release. I am fine with whatever makes the fix correct. Thanks for handling it.

giuseppe:
@mosskappa does the last version make sense to you?

mosskappa:
Yes, the current latest version makes sense to me. I checked the diff and ran a local build plus a minimal FUSE repro for both open(O_TRUNC) and truncate(1); both cases clear SUID/SGID to 0777. The remaining release failure looks like snapshot.ubuntu.com returning 503 for cross packages, not related to the fix logic.

mosskappa:
I also checked the latest head after the retry. It still looks good to me; all checks are green now, and my local build/minimal repro still passes on f9fb3c0.

